PR #79, opened 2015-09-03
Remove svg web fonts

zachleat:
See https://twitter.com/bram_stein/status/506812176212824066

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,6 @@ Download the entire web font archive at this link:
 
 Or select the fonts that you need in subdirectories of the build directory:
 
-- [Download .svg fonts](https://github.com/chrissimpkins/Hack/tree/master/build/svg)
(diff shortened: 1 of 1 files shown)


chrissimpkins:
Thanks for submitting this Zach.  I opened this up for further comment in #88.  Would it be possible to provide further background and justification there?  Can we assume that there is broad use of these versions of the browsers at this point?  The fact that Typekit entirely removed support seems to argue that this will be ok.

zeyadj:
LeavE a comment

chrissimpkins:
@zachleat We eventually did this :)  You were correct

zachleat:
ah, thanks! Sorry I forgot to respond :(

chrissimpkins:
No worries at all.  Just feedback so that you know that I didn't forget :)

PR #110, opened 2015-09-18
License fixes

alerque:


chrissimpkins:
Thanks Caleb. I need to look into this further because DejaVu does not include the license in their own releases of Sans Mono and the suggest that the Arev changes have been merged into their own fonts and released under either the Arev license or in the public domain.  The license in each build suggests the latter.  If this does need to be included, then I will need to update the license display in the builds as well.

alerque:
@chrissimpkins Gotcha. I was going off the fact that they did include the Arev copyright on [this license page](http://dejavu-fonts.org/wiki/License). A look at the Arev project also suggested those glyphs were under a separate license. It's possible that the author also released them to public domain to be fully compatible for inclusion in DejaVu, but it is unclear what "either" means in the context you linked. It could mean some glyphs under one license, some under another or it could mean all glyphs were dual licensed. In the latter case you wouldn't need to mention them at all.

If that can be confirmed from either the Arev or DejaVu side then I'd be happy to drop most of this and redo the PR to just include honorable mention for the sake of courtesy. At that point including the license muddies the waters worse than they need to be.

alerque:
@chrissimpkins I'm re-reading line you [linked to regarding Arev copyright](http://dejavu-fonts.org/wiki/Bitstream_Vera_derivatives#Arev_Fonts), and I think my understanding of it is different than yours. Whether the author actually meant it like this or not should be confirmed, but the page itself is a list of fonts that are derived from Bitstream Vera, not from DejaVu. As such the "importing" of glyphs is relative to that baseline. When it talks about Arev there are two sets of things imported. The Arev is also based on Bitstream Vera but includes the previous (public domain) additions of the DejaVu project as well as its own (copyright) additions. Ergo the "either" wording.

On the other hand the [main license page for DejaVu](http://dejavu-fonts.org/wiki/License) states much less ambiguously that the glyphs imported from Arev are under that font's copyright.

That's my reading anyway. Again I don't think this can be resolved without a confirmation form one or the other of those two upstream parties to know if what was backported from Arev to DejaVu was made public domain or not. Either way using them looks like it's fair game, but what licenses need to be patched on is pretty confusing.

chrissimpkins:
The open source licensing for fonts is somewhat of a mess and the SIL OFL was intended to address that issue in order to bring them under one umbrella.  As I understand it from the SIL OFL FAQ, the Bitstream Vera license was used as source material for the development of the SIL OFL.  They are very similar and the Arev license is a slight modification of the Bitstream Vera license that serves to protect the reserved font names specified in the Arev license.  I don't read any additional restrictions in the license and we have not violated this reserved font name issue.

The problem here is that we are bridging the pre- and post-SIL OFL era of font licensing and it is leading to a multi-license stack which I suspect the original developers never intended.   Most new libre OS typefaces simply use the SIL OFL and one of the questions that I have for SIL is that I believe that derivative projects should simply add new reserved font names to the copyright statements at the top of the license.   The wording of their license is that only the reserved font name displayed to the user (in this project that would be Hack) is considered reserved if I read it correctly.  I am discussing this with a representative from SIL and attempted to contact Monotype (which merged the former Bitstream) to clarify issues on the Bitstream end a bit.  I will try to reach Tavmjong or a representative from the DejaVu group to see if they can address the issue as well.

chrissimpkins:
I reached out to DejaVu project founder Štěpán Roh (@srcx).  It seems that he maintains the Github repository where the project was transitioned (from Sourceforge) earlier this year.  Will update here.

chrissimpkins:
Stepan got back in touch with me and said that Danilo Segan did the Cyrillics for the DejaVu Sans Mono set, pointed to this commit: 

https://github.com/dejavu-fonts/dejavu-fonts/commit/731e9191ad184cc9f81efb45df6fd22725db6400 

It doesn't look like it will be necessary to include this PR or the Arev license.  I think that the current licensing is ok.  Segan is included as one of the DejaVu contributors on our contributors list.

alerque:
@chrissimpkins Savvy. A little poking around seems to confirm that the mono glyphs do have different sources than the other faces and it looks like Arev isn't involved. I'm going to shut down this PR as it is not useful in its current form given what we know about the glyph origins.

chrissimpkins:
Thank you very much for working on this (and for all of your other contributons) Caleb.  I really appreciate it.

Issue #21, opened 2015-08-19
Esperanto support


cdlm:
Esperanto needs just a dozen letters with diacritics: ĉĈĝĜĥĤĵĴŝŜŭŬ (that's `cC gG hH jJ sS` with circumflex, and `uU` with breve)

chrissimpkins:
Will add these and get a version out to you to have a look at.  Hopefully later this week.  Thanks much for the suggestion!

chrissimpkins:
Unicode reference for the development of this set:

Unicode table for Esperanto character set 

http://unicode-table.com/en/sets/esperanto/

chrissimpkins:
Can you let me know what platform you are on?  I am going to prepare some testing builds with these glyphs.

cdlm:
I'm on OS X 10.10.5.

chrissimpkins:
thanks. will let you know when a version is ready to try

chrissimpkins:
Queued for next release

chrissimpkins:
@cdlm Damien, do you mind having a look at the images in PR #118 to see if the positions of the diacritic marks look appropriate on all of these glyphs?  We will create a test build that you can try before we release this if you are available/willing to help us test them.  It would be helpful to have feedback from someone who uses the characters.

chrissimpkins:
From @cdlm : 



First of all, uppercase and lowercase are used as you'd expect in english. However, the accented letters are distinct and sorted just after their non-accented variant in the alphabet. There is no q, w, x, or y. Letters are always pronounced the same (the accented ones are different letters, so they have a different sound, even though it's related to the non-accented one. I can't say for sure, but I doubt there is any word with two adjacent diacritics…

Ĥ and Û are the rarest ones. Ĥ is the greek khi / spanish jota / german hard H as in Bach or in LaTeX; my dictionary has half a page of words starting with Ĥ, and words which originally used it are now orthographied with k instead, wikipedia has a nice explanation and examples.

eĥo (echo), ĥaoso (chaos), Liĥtenŝtejno (Liechtenstein), mariĥuano :)

Ŭ corresponds to english W… my dictionary has only five entries starting with Ŭ, most of them onomatopoeia or transliterations (ŭato means watt, but can also be written vatto). Otherwise, I've only seen used after A or E, as part of a diphtong.

Ŭaŝingtono (Washington, but Vaŝingtono is also used)
Eŭropo (Europe)
eŭfemismo (euphemism)
Aŭgusto (August)
[...]

PR #118, opened 2015-10-04
Add missing Esperanto characters

jublo:
ĈĜĤĴŜŬĉĝĥĵŝŭ and Spesmilo ₷
See #21.


jublo:
bildschirmfoto 2015-10-04 um 14 41 39
bildschirmfoto 2015-10-04 um 14 41 48
bildschirmfoto 2015-10-04 um 14 41 55

cdlm:
Placement is correct from the language point of view, but aesthetically some circumflexes look off-centered on top (mostly the ĉ and Ĝ).

The ĥ is a hard one—but it's a hardly ever used letter. My dictionary centers the circumflex on top of the stem, but other fonts place it more centered on the letter as a whole, on just on the arch like you did. I checked Gentium and actually Gentium Basic and Gentium Plus don't agree on this. In Ubuntu Mono, the circumflex eats a bit of the ascender so that it is still on it, but does not overhang on the left, and the ĵĴ center the circumflex on the serif, which works well. I guess it's a matter of taste and consistency with the rest of the font…

chrissimpkins:
@cdlm thanks for this feedback.  would it be possible to give us several example words that we can use to test the questionable glyphs?  also, are there any words where you would see two glyphs with diacritics adjacent to each other?  This will help us align them and make a decision about style.

cdlm:
First of all, uppercase and lowercase are used as you'd expect in english. However, the accented letters are distinct and sorted just after their non-accented variant in the alphabet. There is no q, w, x, or y. Letters are always pronounced the same (the accented ones are different letters, so they have a different sound, even though it's related to the non-accented one. I can't say for sure, but I doubt there is any word with two adjacent diacritics…

Ĥ and Û are the rarest ones. Ĥ is the greek khi / spanish jota / german hard H as in Bach or in LaTeX; my dictionary has half a page of words starting with Ĥ, and words which originally used it are now orthographied with k instead, [wikipedia has a nice explanation and examples](https://en.wikipedia.org/wiki/Ĥ).

eĥo (echo), ĥaoso (chaos), Liĥtenŝtejno (Liechtenstein), mariĥuano :)

Ŭ corresponds to english W… my dictionary has only *five* entries starting with Ŭ, most of them onomatopoeia or transliterations (ŭato means watt, but can also be written vatto). Otherwise, I've only seen used after A or E, as part of a diphtong.

Ŭaŝingtono (Washington, but Vaŝingtono is also used)
Eŭropo (Europe)
eŭfemismo (euphemism)
Aŭgusto (August)
ĉirkaŭ (around)
antaŭ (before)
antaŭulo (ascendant)
antaŭen (forward)
antaŭeniĝi (to progress)

Ĵ sounds like french J
Ĵuraso
ĵeti (to throw)
ĵurnalo (journal)
belaĵo (a beauty)

Ĝ sounds like english J
ĝi (it, this)
ĝenro (gender)
grandiĝi (to grow, become taller)
manĝi (to eat)
averaĝo (average)

Ĉ sounds like tsh
ĉu ne? (isn't it?)
ĉio (this thing)
aĉeti (to buy)
ĉeno (chain)
ĉefinĝeniero (chief engineer)

Ŝ sounds like sh
ŝi (she)
artiŝoko (artichoke)
ŝovi (to push, to make sth slide)
ŝelako (shellac)
ŝnuro (rope)

I'm still a beginner at the language so I still need to cheat using dictionaries and google translate to find my vocabulary. Maybe grepping through a word list would find more interesting examples…

schierlm:
FTR: There are Esperanto words with two adjacent diacritics. In most cases either the first one is ŭ part of an aŭ or eŭ diphtong (for example antaŭĝoji [to eagerly anticipate] or kaŭĉuko [rubber] or it is a compound word like the (old-fashioned) vojaĝĉeko [traveller's cheque].

And of course, for testing, the (in)famous Eĥoŝanĝo ĉiuĵaŭde [echo change every Thursday] should not be missing (although it does not contain any adjacent diacritics), which is nonsense, but not more nonsense than brown foxes jumping over lazy dogs :)

chrissimpkins:
@cdlm & @schierlm Thanks to both of you for all of these examples. This will give us plenty to work with. 

> but not more nonsense than brown foxes jumping over lazy dogs

Ha!  True, true.

I copied this information to the original issue report thread so that we can continue the discussion there.  This will be merged as we begin work on these glyphs.

cdlm:
Duh, why didn't I think of pangrams! I just found those two, and one even has adjacent diacritics :)

- Eble ĉiu kvazaŭ-deca fuŝĥoraĵo ĝojigos homtipon.
*Maybe every quasi-fitting bungle-choir makes a human type happy.*

- Laŭ Ludoviko Zamenhof bongustas freŝa ĉeĥa manĝaĵo kun spicoj.
*According to Ludwig Zamenhof, fresh Czech food with spices tastes good.*

chrissimpkins:
Great, thank you!

chrissimpkins:
@jublo Do you want to submit this PR as ttf files or as UFO in a `source/development/jublo` directory?  If these ttf are ready to go, I'll start working on them.

jublo:
@chrissimpkins Seems this PR is now superseded by the commits in https://github.com/chrissimpkins/Hack/tree/jublonet-issue/21-esperanto.

PR #151, opened 2015-11-06
Update installation instructions

alerque:
This rewrite introduces basic installation instructions for various platforms
and encourages usage of a few package managers that have recipes for Hack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,10 +75,16 @@ Most Linux and BSD systems can handle either [TTF][ttf_latest] or [OTF][otf_late
 
         $ yaourt -S ttf-hack
 
-* Fedora / CentOS: install from [copr](https://copr.fedoraproject.org/coprs/heliocastro/hack-fonts/):
+* Fedora / CentOS: install from [copr](https://copr.fedoraproject.org/coprs/heliocastro/hack-fonts/). For Fedora >= 23:
 
+        $ dnf install dnf-plugins-core
         $ dnf copr enable heliocastro/hack-fonts
         $ dnf install hack-fonts
+
+  For Fedora <= 22 and CentOS <= 7:
+
+        $ yum install yum-plugin-copr
+        $ yum copr enable heliocastro/hack-fonts
 
 * Ubuntu: install either [fonts-hack-ttf](http://packages.ubuntu.com/xenial/fonts-hack-ttf) or [fonts-hack-otf](http://packages.ubuntu.com/xenial/fonts-hack-otf) for xenial or later:
 
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 - [About](https://github.com/chrissimpkins/Hack#about)
 - [Typeface Data](https://github.com/chrissimpkins/Hack#data)
 - [Type Specimen](http://chrissimpkins.github.io/Hack/font-specimen.html)
-- [Desktop Usage](https://github.com/chrissimpkins/Hack#desktop-usage)
+- [Desktop Installation](https://github.com/chrissimpkins/Hack#desktop-installation)
 - [Web Font Usage](https://github.com/chrissimpkins/Hack#webfont-usage)
 	- [Hack by CDN](https://github.com/chrissimpkins/Hack#hack-by-cdn)
 	- [Self-Hosted Font Files](https://github.com/chrissimpkins/Hack#host-hack-font-files-on-your-server)
@@ -51,21 +51,42 @@ The font binaries are released under a license that permits unlimited print, des
 Click the image for the full type specimen.
 
 
-### Desktop Usage
+### Desktop Installation
 
-#### Recommended Binaries for OS X Users
+Hack is available for download in either [OTF][otf_latest] or [TTF][ttf_latest] formats. Which is best for your system and how best to install these files depends on your operating system. Because Hack is under active development and updates are released frequently we highly recommended using a package manager or other auto-updating utility if possible.
 
-The .otf or .ttf files may be used on OS X platforms.
+#### OS X
 
-- [Download .otf fonts](https://github.com/chrissimpkins/Hack/releases/download/v2.018/Hack-v2_018-otf.zip)
-- [Download .ttf fonts](https://github.com/chrissimpkins/Hack/releases/download/v2.018/Hack-v2_018-ttf.zip)
+The easiest way to install and update Hack on OS is is to use the [Homebrew](http://brew.sh/) package manager which you may already have on your system. Managing fonts with it is as easy as:
 
-#### Recommended Binaries for Linux and Windows Users
+    $ brew cask install caskroom/fonts/font-hack
 
-The .ttf files are recommended for screen use on Linux and Windows platforms.
+To install the font manually you may use either the [OTF][otf_latest] or [TTF][ttf_latest] formats. Download the zip file and extract it. Double clicking each of the font files will open a preview in [Font Book](https://support.apple.com/en-us/HT201749) and the "Install Font" button will copy the font to the correct system location.
 
-- [Download .ttf fonts](https://github.com/chrissimpkins/Hack/releases/download/v2.018/Hack-v2_018-ttf.zip)
+#### Windows
 
+On Windows the [TTF][ttf_latest] format files are recommended. Download the zip, extract the files, and double click on them to open them in the font previewer. Clicking the "Install" button will then copy them to the correct place on your system.
+
+#### Linux / BSD
+
+Most Linux and BSD systems can handle either [OTF][otf_latest] or [TTF][ttf_latest] format fonts. If your distro's package manager has packages for Hack that is the preferred method:
+
+* Arch Linux: install either [otf-hack](https://aur.archlinux.org/packages/otf-hack) or [ttf-hack](https://aur.archlinux.org/packages/ttf-hack) from the AUR either manually or using an AUR helper:
+
+        $ yaourt -S otf-hack
+
+* Fedora / CentOS: install from [copr](https://copr.fedoraproject.org/coprs/heliocastro/hack-fonts/):
+
+        $ dnf copr enable heliocastro/hack-fonts
+        $ dnf install hack-fonts
+
+* Ubuntu: install either [fonts-hack-otf](http://packages.ubuntu.com/xenial/fonts-hack-otf) or [fonts-hack-ttf](http://packages.ubuntu.com/xenial/fonts-hack-ttf) for xenial or later:
+
+        $ apt-get install fonts-hack-otf
+
+   For older systems either manually download one of those packages or see the [manual install instructions](https://wiki.ubuntu.com/Fonts).
+
+For other systems, check for packages using your distro's package manager or download your preferred format and copy the font files to either your system font folder (often /usr/share/fonts/) or user font folder (often ~/.local/share/fonts/).
 
 ### Webfont Usage
 
@@ -308,3 +329,6 @@ Bitstream Vera Sans Mono Copyright 2003 Bitstream, Inc. with Reserved Font Names
 BITSTREAM VERA LICENSE
 
 The full text of these licenses is available in [LICENSE.md](https://github.com/chrissimpkins/Hack/blob/master/LICENSE.md)
+
+  [otf_latest]: https://github.com/chrissimpkins/Hack/releases/download/v2.018/Hack-v2_018-otf.zip
+  [ttf_latest]: https://github.com/chrissimpkins/Hack/releases/download/v2.018/Hack-v2_018-ttf.zip
--- a/README.md
+++ b/README.md
@@ -69,20 +69,20 @@ On Windows the [TTF][ttf_latest] format files are recommended. Download the zip,
 
 #### Linux / BSD
 
-Most Linux and BSD systems can handle either [OTF][otf_latest] or [TTF][ttf_latest] format fonts. If your distro's package manager has packages for Hack that is the preferred method:
+Most Linux and BSD systems can handle either [TTF][ttf_latest] or [OTF][otf_latest] format fonts. If your distro's package manager has packages for Hack that is the preferred method:
 
-* Arch Linux: install either [otf-hack](https://aur.archlinux.org/packages/otf-hack) or [ttf-hack](https://aur.archlinux.org/packages/ttf-hack) from the AUR either manually or using an AUR helper:
+* Arch Linux: install either [ttf-hack](https://aur.archlinux.org/packages/ttf-hack) or [otf-hack](https://aur.archlinux.org/packages/otf-hack) from the AUR either manually or using an AUR helper:
 
-        $ yaourt -S otf-hack
+        $ yaourt -S ttf-hack
 
 * Fedora / CentOS: install from [copr](https://copr.fedoraproject.org/coprs/heliocastro/hack-fonts/):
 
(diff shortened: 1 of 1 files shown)


burodepeper:
Looks good! Quick question: I'm unfamiliar with Homebrew, does it allow automatic updates?

chrissimpkins:
Thanks for adding these instructions Caleb.  Can you confirm that all of these are current with our current release (or how long they typically take to update)?   Also, any reason that you are listing the otf files for Linux users?  My experience based upon past issue reports and discussions is that the ttf builds work better on Linux.

alerque:
@burodepeper Yes, Homebrew handles auto-updating to the latest versions of formula. This is one of the major advantages of using a package manage. Once you install a package on your system it will keep it up to date for you along with the rest of your system. From the standpoint of something like a font that people are likely to install and forget about this means that any time they update anything _else_ on the system they are likely to get the latest updates for your package too. In the case of Homebrew, updating everything on the system including any fonts to their latest versions is:

    brew update && brew upgrade --all

alerque:
@chrissimpkins The Homebrew and Arch Linux packages are at 2.018. All the other systems are behind to one degree or another. Ubuntu is at 2.015, the Fedora/CentOS one is worse (and I've poked the maintainer about it). This is expected however as Hack releases have been fast and furious and most distros have at lease a few days if not weeks lead time for new packages to come down the pipeline. The way most distro release cycles work, people are unlikely to get the latest build the moment it's released; the trade off is that they _will_ get all updates at some point even after they've stopped thinking about it.

As for OTF vs. TTF, I think most of the issues Hack has had with OTF have been errors in properly building the font and I'm hoping those will get resolved with time. As it stands I don't know of any current issues with the OTF version and am not sure why TTF would be a better choice on any *nix system. OTF should be able to handle anything that TTF can including the same curves and tables from a TTF build _or_ more advanced ones. I typically prefer OTF across the board as I work with typesetting software that makes use of the more advanced OpenType features whenever present. Defaulting to the OTF version should allow users to end up with the best feature set if/when Hack includes optimizations or features above and beyond what TTF can produce.

jublo:
@alerque Can we set up the Homebrew Cask to automatically pick up whatever latest tagged release we have? Or how do you plan to keep the Ruby file updated?

jublo:
@alerque @chrissimpkins Also, how do you guys think about the source-foundry becoming the maintainer for the packages referred in this PR docs amendment?

chrissimpkins:
I definitely agree with those comments but the hints are much better at this stage in the ttf than the otf builds.   I believe that we have an open issue report from you about the lowercase u that we have yet to address in the otf files if I recall correctly.  All hinting focus to date has been on the  ttf end and I would be a bit hesitant to suggest otf files for any non OS X user (where the hints are not relevant).

alerque:
@jublo It doesn't work like that\*. Would you want some utility pulling software from thousands of random sources on the net and running it on your machine without some vetting process? All it would take is one compromised project anywhere and chaos unleashed.

Most package managers work on a similar process, but Homebrew's system is pretty basic (and effective). It doesn't take much time but somebody has to pull the trigger and actually update each formula. This means checking the upstream site to make sure it's the latest release and actually legit, downloading the source, generating a new checksum, and making sure there aren't any changes to the install process. For fonts and established projects with standard installations this is pretty easy, a matter of minutes. For some immature (or unorthodox) projects this can mean a bit of work on the part of a package maintainer for each release.

Homebrew formulas are updated using git and anybody on Github can submit updates. My experience is PR's get reviewed and usually accepted within 12 hours.

\* <sub>Technically there is a way to wire up git repos to `--HEAD` install flags, but this only works for brew formulas, not casks. It also means you can't update automatically.</sub>

alerque:
@jublo Again that just isn't how this works.

Most distros do their own package management. In the Linux world that is what a distro is: a team that assembles and maintains known-working combinations of software. You typically don't have upstream projects maintaining their own packages inside distros—that's left to people that care about the distro. There are just too many decisions about where things go and how they should get there to hoist in in from the outside anyway. Many upstream projects have tried to make their software available for \* rather than convincing the distros to pick it up and inevitably this ends up being inept and ineffective.

For distros where nobody close to the project has picked up a package, there are usually ways to add third party repositories. The trick is these usually have to be compiled on the system they are targeting. You can't just go making them willy nilly for systems you don't have running. Also it takes some expertise in the target systems to do this right.

For Homebrew (which is a little different not being the OS but itself a third party tool) the formula repository is owned/maintained by the Homebrew team, but it software is typically updated by community contributed PR's from whoever cares. You can submit updates to Hack as well as I can.

chrissimpkins:
I have much to say on this issue.  Be in touch in a bit. Tied up at the moment.

chrissimpkins:
I fully support these user driven platform-dependent packages and am all for users taking the initiative to redistribute the fonts on their own.  In my mind, this is both useful and necessary from the vantage of our project.  Even if was not, anyone is within licensing assurances to do so (with caveats-see below).  I was not aware that there are as many active packages out there.  I was involved as @legovini developed the Debian package because we came across some Debian specific rendering issues that needed to be addressed before release on Debian.  I know how they are building the packages and we have confirmed that the fonts will display for the users in the same way that they would if they installed directly from our repository downloads.  The rest of these packages were developed without our involvement and knowledge.  I can't say that I know whether they are attempting to build from source, providing modifications to the hinting, or otherwise altering the fonts from our repository releases before they push them to their users.  There are a number of distro-specific issues that need to be addressed in these Linux package releases, from the way that they build the project to licensing issues, and there are involved distro-specific conversations that are held before these packages are approved for release.  I think that this is best left in the hands of people who use these distros and have some level of expertise in the release system and licensing regulations for their own distribution.  

This raises a few issues for this PR which will require a bit of work on our part and more discussion before we merge this as it suggests that we are officially supporting the packages as the proper release and update approach for users on these platforms.

To begin with, we will need to look into how the package maintainers have decided to release the font files.  If they are pulling directly from our repository build files without further modifications, then I am all for recommending this release/update approach to users.  If they are using a different build tool chain, modifying the hints, or otherwise altering the releases from the manner that they are released on this repository, then they are redistributing the fonts as "Hack" outside of the conditions of the license and we cannot promote this.  The reserved font name regulations are there to provide quality assurance to users that they are receiving and viewing builds that represent the decisions of the project team, including the core development team and all project collaborators.  SIL provides guidance on what is considered a "modification" to fonts.  Changes as minor as the addition or removal of a single glyph, a change in hinting, or creation of web fonts from other release formats qualifies as "modification".   Our license has been modeled after the SIL Open Font License with a great deal of input from a member of the SIL team.   When decisions about how the fonts will appear to users are made outside of this context, the license states that they should not carry the "Hack" name.  If this takes place, it makes it impossible for the project to assure that users are receiving the fonts as we intend and to support the releases (and you can be sure that they will show up as issue reports on our repository when there are problems).  

I think that we should reach out to these package maintainers and create a simple communication channel that they can use to discuss the above issues as they develop and maintain their respective packages.  @legovini has joined us in a Slack channel dedicated to redistribution and licensing issues for the typeface.  If other maintainers are interested in doing the same and this is the best approach to discuss these issues, they are welcome.  Knowledge of who these maintainers are and how they approach the redistribution of our fonts will go a long way to recommending their use on our repository.

As for Caleb's suggestion to use these rather than developing a separate tool, I completely agree that this is a tremendous benefit to attach font updates to a user's `brew update && brew upgrade --all` or `sudo apt-get update && sudo apt-get upgrade`.  This is simple, does not require any font-specific upgrade involvement, and is ideal for platforms where this is supported (currently OS X with homebrew and Linux with the distro-specific package managers).  We still fall short on support for Windows users and this is an area where I do feel that a tool that simplifies the upgrade process would be ideal.  I don't know what form that tool would take at the moment and am very open to suggestions.

alerque:
I'm actually rather surprised that there aren't more developed ways of distributing fronts to Windoze users. However on reflection it seems like the main roadblock is that most fonts are ① not freely distributable and ② typically not actively developed. Develop to a final/initial release then make users pay to download and install manually once seems to be the norm.

chrissimpkins:
I am not aware of any other typeface that releases on a similar schedule (if at all) :)

alerque:
@chrissimpkins No licensing issues are involved here. I don't know of any fonts in any package manager than are built from source as part of the process. The packages just automate what a user would be doing manually: downloading the release tarball, extracting in to the right place, triggering a cache update, etc.

alerque:
Can you send me a Slack invite? _«insert mumbling about Slack vs. Open Source»_

chrissimpkins:
@alerque sent!

chrissimpkins:
And to continue this discussion, there was the initial intent on the part of the Debian crew to build from source, hint themselves, and then release the files.  This is why I point this out.  Once we have a build toolchain in place that does not rely on font editors, this will be possible.  Currently, it is not.

heliocastro:
Fedora and Centos 7 copr is up to date now. Centos 6 and 5 will be in the tomorrow update.
[]'s

jublo:
Great to hear, @heliocastro!

heliocastro:
An quick update:
Is easy on Fedora side

- For Fedora < 23 and Centos 7, 6 and 5
yum install yum-plugin-copr
yum copr enable heliocastro/hack-fonts

- For Fedora 23 and beyond
dnf install dnf-plugins-core
dnf copr enable heliocastro/hack-fonts

chrissimpkins:
@heliocastro :+1: thank you!!

chrissimpkins:
@heliocastro is there a good approach to communicate about updates on our end or is this an automated pull from the repository?

alerque:
Thanks @heliocastro! I'll integrate those instructions into the README updates in this PR.

@chrissimpkins I don't know about Copr, but for the AUR packages there is a "flag as out of date" link on each package that notifies the maintainer that there is an update. We should probably assemble a post-release checklist of buttons to push and people to poke for package updates.

- [ ] AUR: flag ttf-hack and otf-hack packages out of date
- [ ] Homebrew: send PR to caskroom/fonts
- [ ] etc.

chrissimpkins:
@alerque sounds great.  let's put together a poke plan when you have a chance

heliocastro:
On copr we don't have the out of date flag, only on fedora official repo, but what i usually do is subscribe news and ping back on the new sources.
I just missed to do that on github, but i don't know if can do that as rss feed.

alerque:
@heliocastro You can get a feed of the releases page on Github by appending .atom to the URL: https://github.com/chrissimpkins/Hack/releases.atom

jublo:
@alerque Thanks for extracting the links of the current versions to endnotes. This will make them easier to update.

README.md
```diff
@@ -51,20 +51,48 @@ The font binaries are released under a license that permits unlimited print, des
...
-- [Download .ttf fonts](https://github.com/chrissimpkins/Hack/releases/download/v2.018/Hack-v2_018-ttf.zip)
+#### Windows
+
+On Windows the [TTF][ttf_latest] format files are recommended. Download the zip, extract the files, and double click on them to open them in the font previewer. Clicking the "Install" button will then copy them to the correct place on your system.
+
+#### Linux / BSD
+
+Most Linux and BSD systems can handle either [TTF][ttf_latest] or [OTF][otf_latest] format fonts. If your distro's package manager has packages for Hack that is the preferred method:
+
+* Arch Linux: install either [ttf-hack](https://aur.archlinux.org/packages/ttf-hack) or [otf-hack](https://aur.archlinux.org/packages/otf-hack) from the AUR either manually or using the AUR helper of your choice:
```
mmlb: should probably favor ttf-hack. Also, ttf-hack is now out of aur and in community, so ISTM `pacman` should be the tool of choice here.

chrissimpkins:
@mmlb 

Is this the correct install command?

`pacman -S ttf-hack`

This PR (with some modifications) will be merged with the upcoming release.

mmlb:
@chrissimpkins yup

chrissimpkins:
@mmlb got it.  will make sure the new information is included. ty!

chrissimpkins:
@mmlb do you need sudo for font installs?

mmlb:
`pacman` will install the package into the system and the font to the system directories so in this case yes. But generally no, the fonts can be install into a user's `$HOME` dir and updating the font cache.

```
cp $font ~/.local/share/fonts
fc-cache
```

chrissimpkins:
@mmlb Got it.  (1) Yes it is necessary; (2) Not necessary for documentation.

Lastly (I think...), does pacman remove outdated font files and update the font cache as part of this install?  There is no need to ask users to manually remove outdated fonts or modify the font caching on the Arch platform?

mmlb:
@chrissimpkins That is correct, the fonts are handled like any other package in the system. Files that are no longer relevant (for multi-file packages) are removed, and post install hooks are called (the hooks for fonts do `fc-cache`).

mmlb:
I can't help but feel like you are mixing system wide fonts and user only fonts, is that so? Just in case to make sure we are on the same page.

`pacman`, which is Arch specific, will install system wide and users only have to install the font. Updates and fc-cache handling is handled.

`~/.local/share/fonts` is *nix generic via `fontconfig`. Users need to manage updating and calling `fc-cache` when the file changes.

chrissimpkins:
I am trying to think through the transition to package manager installs and suspect that some users will convert to use of this Arch specific package manager after a prior manual install of the fonts.  On this distro we can probably assume that most users will have the ability to troubleshoot for themselves, but for my own knowledge and for more comprehensive documentation of this package manager and other Linux PM's out there (and to some degree to head off issue reports that may wind up here for install issues with these package managers that are not project generated), do users need to _manually uninstall_ previously _manually installed_ fonts in `~/.local/share/fonts` before they begin to use the pacman approach (which I am assuming installs in `/usr/share/fonts/`?).  Does Linux in general, or the Arch distro specifically, default to one set of font files over the other if both exist?

Just clarification of the rationale for my question above.  I can look into this in more detail if you don't know off the top of your head.

mmlb:
I see. Did not know off the top of my head (but I looked it up :wink:), so here's the info.

Font paths are completely handled by `fontconfig`, pacman (others too most likely) just put the files in the correct location and call `fc-cache` so that the list can be updated. According to http://cgit.freedesktop.org/fontconfig/tree/fonts.conf.in#n24 the default search order is "system" then "user local" so unless a distro patches that around (seriously wtf though) users should not have a problem with managed fonts.

chrissimpkins:
@mmlb :+1: thank you so much for all of this information! incredibly helpful!

mmlb:
np glad I could help out.

chrissimpkins:
@alerque 
@heliocastro 
@mmlb 

This PR is merged in the `documentation` branch with minor modifications.  Mind having a look at the package manager install section to confirm that it is accurate.  Once it is verified we will merge it into master.  Thanks for doing all of this legwork @alerque.  It looks great.  Sorry for the delay in getting to it.

alerque:
Um...not quite. But I see what happened. I'll submit a fix.

chrissimpkins:
@alerque Merged the update.  Thank you very much Caleb.  Anyone else see other issues?

mmlb:
LGTM. Arch looks good with @alerque's fix, can't really say much about the rest but nothing screams wrong to me.

chrissimpkins:
:+1:  thanks @mmlb. will merge this later today

chrissimpkins:
Merged.  Thanks much to all for these contributions!

PR #180, opened 2016-02-22
added package.json for npmjs.com

### LINKED ISSUE #179: Add Hack to npmjs.com?
Hey, I am not sure you might be interested in this, but if Hack was available from `npmjs.com` people would be able to include it in their `package.json` as a dependency and have it automatically in their Javascript apps.

For example we have an Electron app and now we need to include these fonts in our repository. If available we'd just `install --save` it and it's done.

I can send a Pull request with suggested functionality, with `package.json` if you'd like.

jakubzitny:
resolves #179

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,14 @@
+{
+  "name": "hack-font",
+  "version": "2.0.19",
+  "description": "A typeface designed for source code",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/chrissimpkins/Hack.git"
+  },
+  "license": "SEE LICENSE IN LICENSE.md",
+  "bugs": {
+    "url": "https://github.com/chrissimpkins/Hack/issues"
+  },
+  "homepage": "https://github.com/chrissimpkins/Hack#readme"
+}


jakubzitny:
I'll update the Readme. And for installing the fonts.. I just wanted to have the fonts "inside" my repo, so I can access the CSS files (and the fonts via the font-face). Did you mean installing the fonts in the system "globally"?

chrissimpkins:
I see.  I didn't realize that you do not need to install them system wide to use them in a JS application.

For the README, I think that it would be helpful to have instructions on how to incorporate the fonts as a dependency in a JS/node project and how new releases are addressed with npm, or at the very least a link to npm documentation that explain this.  I am assuming that there must be some documentation out there on how to render glyphs with webfonts using JS?  This won't need to be an exhaustive description.   The goal will be to set a user on the right path who has no experience with this approach.  

This is a great addition to the redistribution of our fonts.  I really appreciate your efforts.  I will get my npm account information to you this evening and will make sure that you are added to our contributors list.  I can merge this today and we can work on the README through another PR or can wait until we have a draft.  Let me know what you would prefer.

jakubzitny:
Sorry I didn't get to this yesterday, hopefully tonight. I think it's better to leave this issue open also for the README part, so the functionality is committed together with the documentation and we don't forget stuff.

The npm publishing process is described in their docs: [Publishing npm packages](https://docs.npmjs.com/getting-started/publishing-npm-packages).

Aand what did you mean by rendering the glyphs in JS?

chrissimpkins:
That sounds fine.  

By rendering the glyphs with JS, I just meant that I wanted to understand how projects such as the one that prompted your PR use the Hack webfonts as the default monospaced face (or body text face, etc).  Is this framework dependent in some way or is there CSS formatting that is common across these frameworks?

allmarkedup:
I'd love to see this merged in once the README is updated. FWIW you don't actually need to publish to NPM to make it installable via NPM, as long as it has a valid package.json then it can be pulled in directory from the Github repo. That is how Adobe are currently handling it for their fonts (e.g. https://github.com/adobe-fonts/source-sans-pro/tree/release).

chrissimpkins:
These appear to be the instructions for their releases:

```
# npm v2.x
npm install --fetch-only git://github.com/adobe-fonts/source-sans-pro.git#release

# npm v3.x
npm install git://github.com/adobe-fonts/source-sans-pro.git#release
```

Can I assume that the install for the Hack fonts would be the following (we use `master` as our release branch rather than an explicit `release` branch):

```
# npm v2.x
npm install --fetch-only git://github.com/chrissimpkins/Hack.git

# npm v3.x
npm install git://github.com/chrissimpkins/Hack.git
```

chrissimpkins:
And for my knowledge, with new releases I assume that I update the package.json version field to the current release as we release.  

https://github.com/chrissimpkins/Hack/pull/180/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R3

Anything else in our repo or outside of it (in npmjs.com) that is necessary with releases?

chrissimpkins:
@jakubzitny My npm account is `simpkins`

allmarkedup:
@chrissimpkins yes those instructions would work fine.

As you say, each release would need the version number in the package.json updating. As typefaces are unlikely to have 'breaking' changes then I'd imaging most releases would increment the 'minor' or 'patch' digits in the version number (i.e. MAJOR.MINOR.PATCH). Note that NPM is very fussy about its version numbers - for example 1.2 won't work, it would need to be 1.2.0 or NPM will complain.

The package.json file is the only thing you need. If you want to publish to NPM you will obviously have to do that seperately after every release is tagged on Github, but as it's a typeface and not JS source code I'm not sure there is much benefit in that. People will be able to install it directly from Github just fine using the instructions above.

chrissimpkins:
@allmarkedup Font versioning is specified to use this odd MAJOR.MINOR system with three digits in the minor portion of the version number.  We were interested in making the change to semantic versioning some time ago, but were advised against this by font gurus who I approached.  It seems to have something to do with versioning/rendering on the operating system side.

We can mock the version string for npm though.  Would it accept a three digit minor version number like so:

`2.019.0`

?

allmarkedup:
@chrissimpkins it will accept three digits, but not with a leading zero, so x.123.x should be ok but not x.012.x. But apart from that caveat, a mocked version number should work just fine.

chrissimpkins:
@allmarkedup Great, thank you very much Mark. We will use the convention as included in the PR then.

chrissimpkins:
@jakubzitny Jakub please weigh in on the discussion above when you have a chance.  I am prepared to move forward with the merge and README changes using the instructions in https://github.com/chrissimpkins/Hack/pull/180#issuecomment-189284485 unless you feel otherwise.

allmarkedup:
@chrissimpkins no problems mate! More than happy to help in any small way I can. I use this typeface a lot and very much appreciate all the effort you've put into it :-)

jakubzitny:
@allmarkedup It is true that it's not really needed on npm. You can still add it via github namespace/repo. I originally thougt that in the future the version on npm might be a bit "ligher" with unneccessary stuff in `.npmignore` and some automatic build scripts. Maybe. But it's also true that it's not really `js` code, and if Adobe has their fonts without npm repository, it might be even better to link it from GitHub.

@chrissimpkins if you use the instructions in the comment you ref'd, we don't really need to upload it to npm and I can delete it from there so it's not misleading and outdated. Also note, that it's good to include a tag or version from git, so once someone installs this in their `package.json` they have it "locked" to a specific version/tag/release. It's also better via `https` I think.

So for the current release it'd suggest ([doc](https://github.com/npm/npm/blob/master/doc/cli/npm-install.md)):
```
npm install --save git+ssh://github.com/jakubzitny/Hack.git#aaf7f399c8cff08a1750f9dbdba5496f7a53041f
```

Anything else I missed?

chrissimpkins:
@jakubzitny I am all for post pull modifications if you are motivated to do that.  It really depends upon what you think the community needs around this redistribution approach.  

We can certainly call that redistribution release approach v2 and go with this for now if the current form is useful.

allmarkedup:
@jakubzitny I certainly don't think there is any problem with publishing it to NPM, as long as @chrissimpkins is ok with the extra step. I'm just not sure it will provide any major benefit over just installing it directly from Github, aside from an more succinct install syntax.

chrissimpkins:
@jakubzitny any reason to favor a hash digest over a simple version tag?

```
$ npm install --save git+ssh://github.com/jakubzitny/Hack.git#2.0.19
```

chrissimpkins:
merged!

chrissimpkins:
@jakubzitny 
@allmarkedup 

Jakub and Mark, can you explain the difference between the npm install with and without the `--save` switch?  As I understand it from the documentation, npm adds the package to your local package.json project dependencies when `--save` is used.  Is this correct and the best set of general instructions for our users?  Is this intended for JS application development or will this work for someone who simply wants to pull the repository using npm and then install the fonts on their system?   I am happy to provide more separate instructions for system installs and application development dependency installs, but I am in somewhat foreign territory in JS world...

jakubzitny:
Hey, yeah the `--save` does exactly that. So it's more for the app developers. 

For just installing it, the `--save` is not needed at all. But for just installing it, people don't really need it from npm, they can use GitHub. Or npm, doesn't really matter. There is even another option for webdevs, and that is bower. Which is more client-side oriented npm-like js repository.

Sorry I didn't get to updating the docs, I still have it in my todo, highlighted :laughing:

chrissimpkins:
> I still have it in my todo, highlighted

Believe me, I completely understand.  I am having a difficult time making a dent in my own to do list... :smile:

chrissimpkins:
@jakubzitny and thanks for the additional information!

chrissimpkins:
For now, I'm sure that we can assume those versed in npm package management will be able to take this into their own hands.  The documentation will be to cover this for the less experienced (including myself!)

PR #198, opened 2016-04-27
Pull Request for changes of the WINDOWS section in README.MD

texhex:
Removed note about C:\windows\fonts (this will just open the FONTS applet for normal users). Updated link to Hack Windows Installer to point to latest version. Added note about restart of computer, just to be sure.


chrissimpkins:
Prefer to link to the repository releases and have them download rather than directly download from the README?  This will be easier to maintain. Adds second click.  I am ok either way.

texhex:
I think the direct link to the latest release in the repo is better because first the user gets a short information what will happen ("Ah, here's the setup so I will click download"), he is able to use that link as a reference to the repo it comes from and finally, one less link you need to care about upon release.

chrissimpkins:
Perfect, I agree.  Merged to master and now live.  Thanks!

texhex:
Link works, and the user will get the release 1.1.2 from ten minutes ago automatically - /latest rules :)

chrissimpkins:
:+1:

PR #207, opened 2016-05-06
Reworked readme.md, about.md, and installation.md

texhex:


README.md
```diff
@@ -1,21 +1,25 @@
 
-# Hack  [![Contributors](https://img.shields.io/badge/contributors-104-orange.svg?style=flat)](https://github.com/chrissimpkins/Hack/blob/master/CONTRIBUTORS.md)
+# Hack  [![Contributors](https://img.shields.io/badge/contributors-104-orange.svg?style=flat)](https://github.com/chrissimpkins/Hack/blob/master/docs/CONTRIBUTORS.md)
 
 ### a typeface designed for source code
 
-No frills. No gimmicks. Hack is hand groomed and optically balanced to be a workhorse face for code. It has deep roots in the libre, open source typeface community and expands upon the contributions of the [Bitstream Vera](https://www.gnome.org/fonts/) &amp; [DejaVu](http://dejavu-fonts.org/wiki/Main_Page) projects.
+No frills. No gimmicks. Hack is hand groomed by Christopher Simpkins and optically balanced to be a workhorse face for code. It has deep roots in the libre, open source typeface community and expands upon the contributions of the [Bitstream Vera](https://www.gnome.org/fonts/) &amp; [DejaVu](http://dejavu-fonts.org/wiki/Main_Page) projects.
```
burodepeper: Unnecessary addition. If something were to change in this paragraph, I'd suggest to remove "No frills. No gimmicks." I might even say that "hand groomed" is not essential to the README. Could be interesting for the real website, less so over here.

README.md
```diff
@@ -1,21 +1,25 @@
...
+* [About Hack](docs/ABOUT.md)
+* [Type specimen](http://chrissimpkins.github.io/Hack/font-specimen.html)
+* [Project website](http://sourcefoundry.org/hack/)
+* [Changelog](CHANGELOG.md)
+* [Contributors](docs/CONTRIBUTORS.md)
+* [License](#user-content-license)
 
 The font binaries are released under a license that permits unlimited print, desktop, and web use for commercial and non-commercial applications. More details in [LICENSE.md](LICENSE.md)
 
-<a href="https://sourcefoundry.org/hack/"><img src="img/hack-specimen-2.png" alt="Hack &mdash; a typeface designed for source code" width="728"></a>
```
burodepeper: Why did you remove the image?

README.md
```diff
@@ -1,21 +1,25 @@
...
-
----
-
 ## Quick installation
 
-We recommend use of the **TrueType** (`.ttf`) version of Hack for most users.
+#### Windows
```
burodepeper: I can't support it with data, but I'm quite sure that Windows users aren't the primary target audience.

README.md
```diff
@@ -24,36 +28,31 @@ We recommend use of the **TrueType** (`.ttf`) version of Hack for most users.
 3. Follow the instructions from your operating system.
 4. Enjoy!
 
-#### Windows
-
-To simplify the installation process on Windows systems, we've created the [Hack Windows Installer](https://github.com/source-foundry/Hack-windows-installer/releases/latest) (`.exe`) which will guide you through the installation process. This installer addresses a number of common rendering issues that occur with font installs on the Windows platform and is the recommended approach on the Windows platform.
+**NOTE ON UPDATING** 
+If you are updating your version of Hack, be sure to remove the previously installed version first to avoid conflicts.
```
burodepeper: No reason to remove the link.

README.md
```diff
@@ -68,38 +67,22 @@ Hack is available in all commonly used web font formats. See [WEBFONT_USAGE.md](
...
 ```
 
-### Additional tools for customization
+#### Additional tools for customization
 
 **font-line** is a tool that lets you easily modify the default line spacing of Hack (20% UPM). See the [font-line repository](https://github.com/source-foundry/font-line) for more details, and be sure to check out the [line-spacing directory of the repository](https://github.com/chrissimpkins/Hack/tree/master/tools/line-spacing) for shell scripts that automate the entire process across several commonly used line spacing defaults.
 
----
-
-## Overview of features
```
burodepeper: I believe these features are pretty important. More relevant to the website, but I believe a subset should be in the README. See conversation between me and Chris on Slack for details.

README.md
```diff
@@ -24,36 +28,31 @@ We recommend use of the **TrueType** (`.ttf`) version of Hack for most users.
...
-To simplify the installation process on Windows systems, we've created the [Hack Windows Installer](https://github.com/source-foundry/Hack-windows-installer/releases/latest) (`.exe`) which will guide you through the installation process. This installer addresses a number of common rendering issues that occur with font installs on the Windows platform and is the recommended approach on the Windows platform.
+**NOTE ON UPDATING** 
+If you are updating your version of Hack, be sure to remove the previously installed version first to avoid conflicts. 
 
-**NOTE ON UPDATING**<br>
-If you are updating your version of Hack, be sure to remove the previously installed version first to avoid conflicts. For more information, see [INSTALLATION.md](docs/INSTALLATION.md).
+#### Manual installation
 
-A [build with CFF curves][otf_latest] (`.otf`) is available for those who prefer this format.
+We recommend use of the [TrueType](ttf_latest) (`.ttf`) version of Hack for most users. A [build with CFF curves][otf_latest] (`.otf`) is available for those who prefer this format. For detailed installation instructions, see [INSTALLATION.md](docs/INSTALLATION.md).
```
burodepeper: "build with CFF curves"? Just mention OpenType, similar to the mention of TrueType.

README.md
```diff
@@ -1,21 +1,25 @@
...
+# Hack  [![Contributors](https://img.shields.io/badge/contributors-104-orange.svg?style=flat)](https://github.com/chrissimpkins/Hack/blob/master/docs/CONTRIBUTORS.md)
 
 ### a typeface designed for source code
 
-No frills. No gimmicks. Hack is hand groomed and optically balanced to be a workhorse face for code. It has deep roots in the libre, open source typeface community and expands upon the contributions of the [Bitstream Vera](https://www.gnome.org/fonts/) &amp; [DejaVu](http://dejavu-fonts.org/wiki/Main_Page) projects.
+No frills. No gimmicks. Hack is hand groomed by Christopher Simpkins and optically balanced to be a workhorse face for code. It has deep roots in the libre, open source typeface community and expands upon the contributions of the [Bitstream Vera](https://www.gnome.org/fonts/) &amp; [DejaVu](http://dejavu-fonts.org/wiki/Main_Page) projects.
 
-[Read more](docs/ABOUT.md) about Hack, visit the [project website](http://sourcefoundry.org/hack/), take a look at a complete [type specimen](http://chrissimpkins.github.io/Hack/font-specimen.html) or browse the [CHANGELOG](CHANGELOG.md).
```
burodepeper: I'm in two minds about the contents of either this paragraph, or the list it has turned into. My issue is that part of it is/are inline links, part are relative links, and part are outside links. Having them as a list seems a bit less intuitive, because that suggests them to be equal to me.

README.md
texhex: This two sentences give a good overview about Hack so I vote for keeping them. But if there is "hand groomed", there should be a "by X".

README.md
texhex: I moved it to ABOUT in order to keep the README short and also because I don't think the specimen alone has enough information value to qualify for the README. If someone really comes here and has no idea how Hack looks, he/she can access ABOUT.MD.

README.md
texhex: This has not been resorted because of primary audience, but because the Mac/Linux section consists of a list and a special note about updating. Reordering keeps the simple stuff on top and the other stuff below. I think it makes reading it simpler.

README.md
texhex: Link moved. See your other comment.

README.md
texhex: OK, overlook that information.

README.md
texhex: I have no idea what "CFF curves" means, this was the exact text I copied from your version of the README so I though this would be important.

README.md
texhex: I personally like lists more, because when you search for something, you just check all the keywords and go directly for what you are looking for. But I agree with you, this is also a personal preference. I have no problem with it when Chris choose to trash the link list.

About the links: The user only cares if clicking the link results in the expected information. He does not care if the target is inline, relative or external.

README.md
chrissimpkins: We can remove the hand groomed comment.  It is unnecessary to have my name included.  I like the optically balanced line.  I think that it is appropriate and it is one of the ways that it differs from Bitstream Vera Sans Mono and DejaVu Sans Mono

README.md
chrissimpkins: I favor leaving at least one of the specimen images on the README.  It is a nice way to provide an introduction to the typeface and it may be best to link to the type specimen page rather than the Hack homepage on the website.

README.md
chrissimpkins: I would either arrange by most common user type (probably OS X, then Linux, then Windows) or alphabetically.  In either case, I agree with David that I would drop Windows down in the order.

README.md
chrissimpkins: otf builds aren't properly referred to as OpenType files.  Both ttf and otf files use OpenType technology.  ttf builds use quadratic Bezier curves, otf builds (CFF outlines) use cubic Bezier curves.  For details:

https://en.wikipedia.org/wiki/OpenType#Description

It's too complex to worry about in our README.  Let's just call them ttf and otf builds.  What we want users to know is the file type that they should use, not the details.

README.md
chrissimpkins: I like having a comprehensive list of available package managers here.  The only reason that it hasn't been more complete is that I didn't know about these packages.  How best to display this?  Should we use a table that includes the 

1. Platform
2. generic command
3. platform specific package manager Hack package name

?

README.md
chrissimpkins: It was originally a list.  I generally format a Table of Contents list at the beginning of long README documents and that is why I included it here.  I am fine either way with this.  I like the paragraph style, I like the list style.  If we go back to a list, I would favor a more comprehensive Contents list that includes the package managers and links to all of the information that is now moved to the docs subdirectory.

burodepeper: If it is indeed a table of contents then a list best suits it. Right now, it isn't, so I think displaying this data as a list is deceptive to a user.

README.md
burodepeper: > Let's just call them ttf and otf builds. What we want users to know is the file type that they should use, not the details.

Seconded.

README.md
burodepeper: Perhaps a list of _which_ package managers have support, and link to a secondary document for specifics?

chrissimpkins: > Perhaps a list of which package managers have support, and link to a secondary document for specifics?

Yeah, that is not a bad idea.  Still begs the question, how best to display the commands no matter where we place them.

chrissimpkins:
Tex, weigh in when you have a chance.  We seem to be closing in on a compromise.

README.md
burodepeper: What's wrong with the way they're displayed right now in INSTALLATION.md? That's good enough for now I reckon.

chrissimpkins: @burodepeper didn't see the former version of the text in the INSTALLATION.md file.  Yes, that is fine.  I agree.  We can add the new packages from @texhex PR.

README.md
texhex: Hand groomed by Chris... line removed, although in pain.

README.md
texhex: Picture added below license in first section.

README.md
texhex: Restored old OS order.

README.md
texhex: OK, changed text of OTF link text.

README.md
texhex: >I believe these features are pretty important.

Added *Overview of features* again.As I understood it, the list of package mangers is OK as is.

README.md
texhex: Removed *Project website* and *Full specimen* external links and included all sections. Created another section called "Resources" that list all the old external lists.

PR #215, opened 2016-06-15
add note for the jetbrains rendering issues

jorgheymans:
@chrissimpkins feel free to reword or reformat , not sure if we should link this paragraph somewhere more prominent given the popularity of the jetbrains toolsuite. I'll leave that up to you :-)


chrissimpkins:
Is the comment about JDK9 still accurate Jorg? I am under the impression that there are still problems there.

jorgheymans:
As far as I can tell yes it solves some issues but apparently not all of them. I think there's still value in putting in the note in the documentation, but yeah it's still a beta jdk so not likely to be used mainstream ytr yet.

chrissimpkins:
Thanks @jorgheymans!

Issue #201, opened 2016-04-28
Trigram for heaven glyph is missing


mike-zueff:
Hello @chrissimpkins,

I use Hack font version 2.020 and vim-airline plug-in. The Unicode glyph for 0x2630 is unfortunately not implemented yet, so I see a rectangle placeholder in the status line. This glyph is a trigram for heaven and should look like the corresponding trigram in the following Wikipedia article:

https://en.wikipedia.org/wiki/Bagua

Could you please implement it (both bold and regular)?

chrissimpkins:
Mind pushing a screenshot of a font that includes it so that we can see it in context within the vim-airline UI?

mike-zueff:
Some screenshots can be found here:

https://github.com/vim-airline/vim-airline/pull/1103

chrissimpkins:
@mike-zueff Thanks Mike! 

For my future reference:

<img width="554" alt="trigram heaven" src="https://cloud.githubusercontent.com/assets/4249591/14887378/8e25ca50-0d23-11e6-9c7d-fa5af76f0e0c.png">

Is it always positioned adjacent to numerals on the airline UI as shown in this image?  Any other uses?

![26deed3c-f102-11e5-8377-b2fb874aa0a6](https://cloud.githubusercontent.com/assets/4249591/14887401/adc11892-0d23-11e6-8a5f-ed85d3b82a15.png)
![e0193c2c-088f-11e6-834f-36a28f0de02d](https://cloud.githubusercontent.com/assets/4249591/14887432/d213c988-0d23-11e6-8270-1ff0327c78cb.png)

mike-zueff:
![2016-04-28-185125_631x137_escrotum](https://cloud.githubusercontent.com/assets/16763520/14892178/b2380dca-0d7a-11e6-947e-4e008b4105f8.png)

There's no any other uses for this glyph in vim-airline. In the most recent default status line configuration for vim-airline, this trigram is always located on the right of the number displaying total lines count in the file. On my screenshot, it's replaced by a missing glyph placeholder.

kanbara:
I also don't have this glyph in Envy Code R for Powerline. Probably should use a more standard character as it's hard enough to get good console fonts patched for powerline. (re: effort)

chrissimpkins:
@kanbara encourage those Powerline and variant developers to quit being so creative ;)

chrissimpkins:
@iamjamestl are you able to rebase your PR on current dev branch?

[...]

PR #236, opened 2017-01-06
Add Yijing mono-, di-, and trigrams (primarily for vim-airline)

iamjamestl:
14 glyphs from 2630-2637 and 268A-268F.  These are based on the glyphs from DejaVu Sans (not mono).  The widths were copied from 268A and 268B in DejaVu Sans Mono Regular, which are the only two of the 14 to exist.

2630, trigram for heaven, consisting of three horizontal lines is used to indicate the line count in vim-airline.  Other symbols my be used to indicate a column layout.

- URxvt
trigram-urxvt
- Details
trigram-heaven

Here is Hack v2.020 with these new glyphs patched in (along with my changes from #234).
[Hack-v2.020-with-ngrams.zip](https://github.com/chrissimpkins/Hack/files/689089/Hack-v2.020-with-ngrams.zip)

Re: issue #201

### DIFF
--- a/source/Hack-Bold.ufo/glyphs/uni2630.glif
+++ b/source/Hack-Bold.ufo/glyphs/uni2630.glif
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<glyph name="uni2630" format="1">
+  <advance width="1233"/>
+  <unicode hex="2630"/>
+  <outline>
+    <component base="uni268A"/>
+    <component base="uni268A" yOffset="646"/>
+    <component base="uni268A" yOffset="1293"/>
+  </outline>
+</glyph>
--- a/source/Hack-Bold.ufo/glyphs/uni2631.glif
+++ b/source/Hack-Bold.ufo/glyphs/uni2631.glif
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<glyph name="uni2631" format="1">
+  <advance width="1233"/>
+  <unicode hex="2631"/>
+  <outline>
+    <component base="uni268A"/>
+    <component base="uni268A" yOffset="646"/>
+    <component base="uni268B" yOffset="1293"/>
+  </outline>
+</glyph>
--- a/source/Hack-Bold.ufo/glyphs/uni2632.glif
+++ b/source/Hack-Bold.ufo/glyphs/uni2632.glif
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<glyph name="uni2632" format="1">
+  <advance width="1233"/>
+  <unicode hex="2632"/>
+  <outline>
+    <component base="uni268A"/>
+    <component base="uni268B" yOffset="646"/>
+    <component base="uni268A" yOffset="1293"/>
+  </outline>
+</glyph>
--- a/source/Hack-Bold.ufo/glyphs/uni2633.glif
+++ b/source/Hack-Bold.ufo/glyphs/uni2633.glif
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<glyph name="uni2633" format="1">
+  <advance width="1233"/>
+  <unicode hex="2633"/>
+  <outline>
+    <component base="uni268A"/>
+    <component base="uni268B" yOffset="646"/>
+    <component base="uni268B" yOffset="1293"/>
+  </outline>
+</glyph>
--- a/source/Hack-Bold.ufo/glyphs/uni2634.glif
+++ b/source/Hack-Bold.ufo/glyphs/uni2634.glif
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<glyph name="uni2634" format="1">
+  <advance width="1233"/>
+  <unicode hex="2634"/>
+  <outline>
+    <component base="uni268B"/>
+    <component base="uni268A" yOffset="646"/>
+    <component base="uni268A" yOffset="1293"/>
+  </outline>
+</glyph>
--- a/source/Hack-Bold.ufo/glyphs/uni2635.glif
+++ b/source/Hack-Bold.ufo/glyphs/uni2635.glif
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<glyph name="uni2635" format="1">
+  <advance width="1233"/>
+  <unicode hex="2635"/>
+  <outline>
+    <component base="uni268B"/>
+    <component base="uni268A" yOffset="646"/>
+    <component base="uni268B" yOffset="1293"/>
+  </outline>
+</glyph>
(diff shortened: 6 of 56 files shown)


chrissimpkins:
Greatly appreciate this work @iamjamestl.  Let's leave this one open to work on / address down the road.

chrissimpkins:
Is this something that you still use and have interest in including in Hack @iamjamestl?

iamjamestl:
Yes, I will rebase this PR tonight and you can decide whether you want to include it or not.

chrissimpkins:
No rush.  Will likely not be able to get to this for a couple of weeks but happy to include it if it is still of interest.

chrissimpkins:
Seems that we have a few issue reports open for these glyphs and this will address all of them.

iamjamestl:
I've rebased this branch on top of the current master according to CONTRIBUTING.md, which is to say, it only contains the `.glif` files.

I think it's worth mentioning again that I copied these glyphs almost directly out of DejaVu Sans.  I don't know anything about Korean language or culture so I am just trusting that DejaVu got it right.

chrissimpkins:
Sorry, I missed that comment about DJV Sans previously.  This is coming from Mono or proportional Sans version of DJV?  If the latter, will you please point me to the license for these glyphs?  I know that the DJVS Mono changes to the Bitstream VSM upstream were committed to the public domain.

iamjamestl:
I based the glyphs on DejaVu Sans (Proportional), which added the glyphs in version 2.4, making them additions to DejaVu, not originally coming from Bitstream Vera.  Thus, according to [the DejaVu license](https://dejavu-fonts.github.io/License.html), the glyphs are in the public domain.

chrissimpkins:
Are you able to confirm that these are not part of the Arev sets that would require the addition of another license to the repository?

Arev license is in the link that you provided above and this is the information that I can find on the Arev fonts:

http://tavmjong.free.fr/FONTS/

iamjamestl:
Before DejaVu, the glyphs came from a Bitstream Vera derivative called Verajja (https://sourceforge.net/p/dejavu/mailman/message/7275795/; note glyph additions 9776-9783, 9866-9871) which the DejaVu project got permission to merge into DejaVu Sans (https://sourceforge.net/p/dejavu/mailman/message/7275522/).

chrissimpkins:
From Stepan Roh of DejaVu team:

```
Hi.

On Sun, 22 Jan 2006, Stepan Roh wrote:

> http://www.aimwell.org/Fonts/fonts.html, search for Verajja (fonts are in 
> 7zip archive)
>
> It is Vera Sans with additional characters for Pali and Vietnamese. Rough 
> comparison of Unicode coverage shows that it has greater coverage for Latin 
> Extended-B, full coverage for Latin Extended Additional and lots of, I am not 
> afraid to say so, very cute symbols.
>
> Fonts are only in TTF and their license is unknown. I will send a mail to the 
> author and see what happens. Do not import glyphs from these fonts until we 
> have a permission. Thank you.

I received a permission to merge it. I don't have time right now to do the 
merge so if there are any volunteers...

If someone has a problem with unpacking the .7z files, let me know and 
I'll write you where to download them unpacked.
```

Thanks James.  Will need to look into this further.  I am assuming permission to merge it means that they were ok to then release these changes to public domain (which itself is a problem and prevents us from releasing through Google Fonts).

iamjamestl:
You do what makes you comfortable, but ultimately, this is my work which I am giving to the Hack project under the terms of the MIT license.  I wrote the XML and I made decisions about how to fit the glyph into the narrower width of Hack.  These glyphs are a derivative work of the glyphs in DejaVu Sans.  The question you should ask is whether the DejaVu license permits me to make a derivative work, and under what license the derivative work must be placed.  I think any fair reading of the DejaVu license, whether the glyphs fall under the Bitstream Vera portion or the public domain portion, permits me to make a derivative work under the license of my choosing, except if the glyphs fall under the Bitstream Vera portion (which I don't think they do), and even then, it only dictates that the Bitstream Vera copyright information and original permission notice be retained, which Hack already does.

chrissimpkins:
> The question you should ask is whether the DejaVu license permits me to make a derivative work, and under what license the derivative work must be placed. 

Yes, this was absolutely the question and I apologize if you took this to mean that I did not feel that you were the author of modifications to the upstream.  Irrespective of whether you developed a derivative or these were direct merges from the upstream designs, the upstream licensing must be maintained for some licenses and my only concern is that I don't completely understand the upstream licensing situation here.  I am guessing that the agreement was that DJV was able to release these glyph under same public domain approach that they used for their other glyphs (something that I do not understand whether applies outside of the DJV Sans Mono fonts because I have not looked into it for other DejaVu fonts).  It is not explicit in the above email message, only that it was ok for DJV to merge and release them based upon second hand information from a DejaVu author who was not the author of this work.   Assuming that they did not modify the license for DejaVu Sans (proportional, not Mono) with the merge of this work, it is implied that this was the agreement from that upstream author.   I have spoken with Stepan about other DJV issues (we were able to track down the actual commits for Cyrillic glyphs in the sets to confirm that the DJV public domain release applied).  I can reach out to him and the author of the merged work to confirm that we are clear here. 

The Google Fonts comment was simply commentary and not intended to suggest that we wouldn't consider it for that reason.  We have other DJVSM glyphs in the sets and the only way that we could work around this issue is to redraw all of them in order to eliminate the public domain portion of the license which would be a ludicrous duplication of efforts already put forth by this upstream team.   Simply to point out that public domain work raises their own set of issues despite appearing "benign" licensing.

I can assure you that these glyphs do not fall under the Bitstream project.  Those sets did not include these expanded character sets.  This would all fall under DejaVu work.

I will let you know what I learn once I dig into this issue.

iamjamestl:
I am withdrawing this patch.  These glyphs have issues.  Namely, current glibc does not know how wide they should be.  [utf8proc](https://github.com/JuliaLang/utf8proc) says they should be double-width, which makes sense given their Korean origin.

```
> cat test.c
#include <stdio.h>
#include <wchar.h>
#include <utf8proc.h>

int main() {
    wint_t c = 0x2630;

    printf("wcwidth returns %d\n", wcwidth(c));
    printf("utf8proc_charwidth returns %d\n", utf8proc_charwidth(c));

    return 0;
}

> ./test
wcwidth returns -1
utf8proc_charwidth returns 2
```

This can cause all sorts of problems in certain apps depending on how they do string width calculation.  tmux, for example, assumes all glyphs with an unknown width are size 1: https://github.com/tmux/tmux/blob/master/utf8.c#L130.  So that works with this MR given I made the glyphs one character wide.  But when you compile tmux with utf8proc (`./configure --enable-utf8proc`), which one needs to do on old operating systems with extremely out-of-date Unicode support, like Solaris 10, then tmux thinks they it should be two characters wide and that causes screen corruption.

For vim-airline users (see #201), I would suggest adding the following configuration to your `.vimrc`:

```vim
if !exists('g:airline_symbols')
    let g:airline_symbols = {}
endif
let g:airline_symbols.maxlinenr = '≡'
```

That will cause airline to use a well-known single-width glyph that is already included in Hack and looks very similar to the trigram:

screenshot_20171116_142405

chrissimpkins:
@iamjamestl amy idea why they went with those glyphs for the airline UI? That seems bound to cause havoc.  Sorry to hear this James.  Thanks for going to all of the work to make them.

PR #250, opened 2017-06-17
Do not recommend fc-cache -s in INSTALLATION.md

jacksgt:
`fc-cache -s` will only scan system-wide font directories, omitting user dirs.
(The line above points out that fonts can be put in a local directory, so this option doesn't make sense)


chrissimpkins:
Thank you for picking this up!  Much appreciated.

For future reference from `fc-cache` docs:

>If directory arguments are not given, fc-cache uses each directory in the current font configuration. Each directory is scanned for font files readable by FreeType. A cache is created which contains properties of each font and the associated filename. This cache is used to speed up application startup when using the fontconfig library.

Issue #262, opened 2017-08-11
Contributing to the Hack typeface - create documentation


chrissimpkins:
- [x] design philosophy
- [x] design optimization targets
- [x] documentation for issue reporters
- [x] documentation for code contributors

chrissimpkins:
initial draft of design philosophy and optimization targets is available in the `contributing-docs` branch: 

https://github.com/source-foundry/Hack/blob/contributing-docs/CONTRIBUTING.md

chrissimpkins:
@burodepeper poke

when you have a chance

burodepeper:
@chrissimpkins I knew I had forgotten something. Will have a look today.

chrissimpkins:
relevant and associated with this discussion https://github.com/source-foundry/Hack/issues/271#issuecomment-323086804

chrissimpkins:
updated with IR and PR documentation in #275

chrissimpkins:
merged to master

PR #272, opened 2017-08-16
Update CONTRIBUTING.md

burodepeper:
/cc @chrissimpkins #262 

I haven't changed very much. I think it is important that this grows over time, but is kept as concise as possible at the same time.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,28 +1,24 @@
-## Hack Typeface Design
+# Hack Typeface Design
 
 Hack is a monospaced typeface that is designed to optimize the display of source code text.
 
+This document outlines the core ideas on which we build to continually improve Hack. Ideally, they serve as a final rationale to decide disputes of whatever nature.
 
-### Primary Design Targets
+## Design Targets
 
-- ASCII glyph set (used to display the body of all source code text)
-- font sizes between 8 - 14 px
-- cross-platform, cross font renderer support on Linux, OS X, and Windows operating systems
-- desktop text editor text displays
-- terminal emulator text displays
-- website text displays
-- print text displays
+Hack is a general purpose typeface for source code. The _needs of the many_ describe our core design targets. Generally, an issue that affects 90% of users, gets a higher priority than one that 'only' affects 10%.
+
+### Core
+
+- **ASCII glyph set**; generally speaking, all source code is limited to ASCII. Content/comments/documentation, on the other hand, often includes non-ASCII characters. The former gets precendence over the latter.
+- Font-sizes between **8-14 px**, line-height >= 1
+- **Cross-platform**, cross font renderer support on Linux, OS X, and Windows operating systems
+- Usage in **common developer scenarios** (on digital displays): text editors, terminals, embedded as webfont, etc.
+
+## Goals, areas of improvement
+
+- **Legibility** - establish differences in the appearance of similar glyph shapes so as to properly identify different Unicode code points
+- **Readability** - glyph shape and spacing optimizations to improve the capacity to read character-character, word-word, and code block-code block combinations in source code
+- **Visual semantics** - establish semantic commonalities for glyphs used in source code text and create common visual designs within these semantic groups
 
 
-### Primary Design Optimizations
-
-- legibility - establish differences in the appearance of similar glyph shapes so as to properly identify different Unicode code points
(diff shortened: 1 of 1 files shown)


burodepeper:
Indeed, Powerline is not a core target. I included terminals, because they're such a crucial part of the developer's workflow. I think we can apply the rules in the document already: ASCII first, Powerline (and others) second.

burodepeper:
I've read it again, and I don't know if it needs to be addressed separately. I think it is clear that the core focus is on ASCII, and once you start naming other variants and extensions, where do you stop?

I do think that the document needs a statement about variations of glyphs, such as a slashed zero, and to what extend that is part of the core effort. I think that we choose a primary variant, and that subsequent variants --while technically ASCII-- are of secondary concern also.

chrissimpkins:
> Indeed, Powerline is not a core target. I included terminals, because they're such a crucial part of the developer's workflow. I think we can apply the rules in the document already: ASCII first, Powerline (and others) second.

Agree. I think you are right.  This is answered with the ASCII glyph set point.

> I do think that the document needs a statement about variations of glyphs, such as a slashed zero, and to what extend that is part of the core effort

This is an excellent point.  My bias is that style modifications with major design adjustments for aesthetic reasons alone belong in forks.  These could be forks that we release ourselves, but forks nonetheless.  Before we commit to that, do you see an area for improvement in the style of any of the ASCII glyphs?  There will be work that is aimed at our optimization goals (including potentially relatively significant shape changes with the visual semantics goal that you are interested in), but is there anything that stands out from an aesthetic standpoint in need of change in your opinion?  I know this is purely subjective but there is a highly subjective component to font selection/use and I am interested in your thoughts.

> I think that we choose a primary variant, and that subsequent variants --while technically ASCII-- are of secondary concern also.

I'm not sure that I agree with this.  My current syntax highlighter is using a combination of italic and regular in the code itself and if I had the time to modify it I would probably add a touch of bold here and there.  I don't think that the amount of work is going to be significant enough that we will need to limit this to a single variant set and my expectation as a user would be at the very least a functional set of consistent optimizations across all of the released variants in the ASCII set.

chrissimpkins:
How do you want to handle merges?  Feel free to merge into `contributing-docs` from my standpoint or I am happy to take care of it.  

Once we settle on these pieces, the next step will be the issue reporting and PR documentation.  Shouldn't be too much trouble and I am happy to get started on it unless you have a strong desire to write the first draft.  Let me know.

burodepeper:
> >I think that we choose a primary variant, and that subsequent variants --while technically ASCII-- are of secondary concern also.

> I'm not sure that I agree with this.

Ah, that's not what I meant :) By variant, I meant a variant on a single glyph, dotted zero vs slashed zero.

> There will be work that is aimed at our optimization goals (including potentially relatively significant shape changes with the visual semantics goal that you are interested in), but is there anything that stands out from an aesthetic standpoint in need of change in your opinion?

Not change related, nothing major at least. I do want to have a thorough look at the alignment of all the punctuation in ASCII, basically everything except for the alphanumeric characters.

> How do you want to handle merges? Feel free to merge into contributing-docs from my standpoint or I am happy to take care of it.

I'll read through it again tomorrow with these last remarks. I'll merge this, and if you could then flesh out the issue reporting and pr docs, I can have a look at that over the weekend.

chrissimpkins:
> Ah, that's not what I meant :) By variant, I meant a variant on a single glyph, dotted zero vs slashed zero.

Ah, yes.  I suggest that we consider variant shapes something for separate forked projects.  Dotted zeroes, slashed zeroes, etc.  The new UFO source + scripted build process will make these forks very simple to make.  All that someone would need to do is drop a .glif file into the sets with a different design and rebuild with the script, then rename the fonts to generate a completely new fork.  If they modify a glyph that is manually hinted, they will need to modify the hinting script.  I was considering a repository to maintain variant glyphs that could be used for exactly this purpose along with a tool that will rename and version the fork.  This would be a very simple approach for someone who wants to create a fork and keep it current with changes in the Hack upstream.

> I'll merge this, and if you could then flesh out the issue reporting and pr docs, I can have a look at that over the weekend.

:+1:  sounds great.  will do.  thanks David

chrissimpkins:
Merged these changes and will continue work on `contributing-docs` branch.  Deleting temporary `burodepeper-patch-1` branch from the remote.

Will add information about glyph variants and approach to them.  Feel free to pick up work on the updated `contributing-docs` branch when you have time.

Issue #269, opened 2017-08-15
Add CI testing for build shell scripts


chrissimpkins:
Add Travis CI testing for the new build.sh shell script that builds Hack fonts from UFO source.

Recommended by @mavit in https://github.com/source-foundry/Hack/issues/227#issuecomment-322526873

Testing tool suggested: https://www.shellcheck.net/

Travis CI implementation: https://github.com/koalaman/shellcheck/wiki/TravisCI

TODO:

- [x] implement shellcheck testing on `build-web.sh` script
- [x] implement shellcheck testing on Werner's ttfautohint install script
- [x] implement shellcheck testing on web font build scripts

mavit:
Looks like this is now implemented on branch `dev-build-scripts-shellCItest-patch`?

chrissimpkins:
Yes and seems to be working well.  I really appreciate the suggestion.  Sorry haven't had a chance to update the issue report yet.  Was on the agenda for this weekend.  

I do have a question for you if you have experience with Travis.  I would prefer to implement this concurrent to the ufolint tests using their matrix build process.  In the branch you mentioned, I have the test in line (i.e. sequential) with our ufo source tests.  Do you know of a way to split this out to a separate matrix 'build' using Travis terminology so that it runs as an isolated test independent from the ufo source tests?  I am aiming to do this with the build.sh test that is a simple execution of the build.sh script with the dependency install flag.  

Note currently failing in that branch because of my failed attempts to install the harfbuzz dependency.  Werner Lemburg contributed a script to take care of these elements and once I have this in place over the weekend we should be passing tests again.

chrissimpkins:
merged into `dev-build-scripts` branch while will remain remote working branch for the build script work

https://github.com/source-foundry/Hack/commit/19a2b9390e49d7db93a90e04e05fe51b3e79b187

chrissimpkins:
In addition to shellcheck tests of build.sh script, the test now includes complete script execution with dependency installs for fontmake, fontTools, and ttfautohint.  Passes tests on Travis as of https://github.com/source-foundry/Hack/commit/19a2b9390e49d7db93a90e04e05fe51b3e79b187

mavit:
> In the branch you mentioned, I have the test in line (i.e. sequential) with our ufo source tests. Do you know of a way to split this out to a separate matrix 'build' using Travis terminology so that it runs as an isolated test independent from the ufo source tests?

[...]

PR #274, opened 2017-08-20
Run lint via parallel make.

chrissimpkins:
Adds makefile for concurrent CI testing approach

Contribution from @mavit


chrissimpkins:
Discussion in https://github.com/source-foundry/Hack/issues/269

chrissimpkins:
Merged into `mavit-make` branch.

Issue #286, opened 2017-08-24
Proposal: Web font builds


chrissimpkins:
woff is supported in all modern browsers.  TTF, EOT, and SVG web fonts are no longer recommended.

Suggest that we build woff (+/- woff2) files **only** for web font releases.

- [x] convert to woff + woff2 only web font build?
- [x] subset web font build - ASCII + Latin-1 + Latin Extended A subset for web fonts
- [x] create web font build scripts - added `build-woff.sh` & `build-woff2.sh` & `build-subsets.sh`
- [x] add make approach for webfont builds (`$ make webfonts`)
- [x] evaluate the glyphs contained in the web font subsets to confirm that defined glyphs are included
- [x] confirm that middle dot glyph is defined in bold italic web font subset.  It is included in full set but not in subset (and is included in all other subsets) see https://github.com/source-foundry/Hack/issues/286#issuecomment-330007092 - the .glif file was missing.  This has been fixed in PR #320

chrissimpkins:
web font build scripting is underway in the `dev-build-scripts` branch.

chrissimpkins:
woff browser support from https://caniuse.com/#feat=woff

![woff](https://user-images.githubusercontent.com/4249591/29731933-735b20c6-89b4-11e7-9928-bbd146bc0f2b.png)

chrissimpkins:
woff2 browser support from https://caniuse.com/#feat=woff2

![woff2](https://user-images.githubusercontent.com/4249591/29731954-82a3c826-89b4-11e7-86b6-4acf8cc6754f.png)

chrissimpkins:
Scripting for ttf to woff build is now available in `build-web.sh` script `dev-build-scripts` branch.  
To install dependencies before the build, use:

```
$ ./build-web.sh --install-dependencies
```

All subsequent builds:

```
$ ./build-web.sh
```

Builds currently land on the path `build/web/*.woff`

chrissimpkins:
Added new `build-woff2.sh` script in `dev-build-scripts` branch.  Builds woff2 web fonts from ttf build files.  I confirmed that there is a significant compression change c/w the woff files.  (~150kb to ~100kb).

We will plan to release both.  

woff2 build adds a new build dependency = https://github.com/source-foundry/woff2

[...]

PR #291, opened 2017-08-29
ASCII + Latin-1 + Latin Extended A + currency + general punctuation subset data added to lib.plist

chrissimpkins:
…for subset builds

@burodepeper If you have some time, can I ask you to double check me on these lists for the web font subset?  They are all identical new `lib.plist` files across each variant so only necessary to examine one set.  I just want to confirm that another set of eyes has a look through to confirm that I didn't insert a misspelling or skip one of the glyphs in the sets.

The only one that *should* be missing is U+017F and that is because it is not included in our sets.

The list requires the glyph "name" not the file path to the .glif file.

The Unicode Standard is here: http://www.unicode.org/Public/6.2.0/charts/CodeCharts.pdf

Thanks

TODO: 

- [x] how to handle currency symbols and general punctuation glyphs (see https://github.com/source-foundry/Hack/pull/291#discussion_r135722204 and https://github.com/source-foundry/Hack/pull/291#discussion_r135726544)
- [x] build test sets to determine final web font file sizes for woff and woff2


chrissimpkins:
had to add the `public.glyphOrder` key back to the lib.plist file. fontmake uses it to determine the glyphs that are built into the font (without subsetting).  The subset list is an array under the key `com.schriftgestaltung.Keep Glyphs`.

source/Hack-Regular.ufo/lib.plist
```diff
@@ -2,3346 +2,324 @@
...
-		<string>uni200A</string>
-		<string>uni202F</string>
-		<string>Abreve</string>
-		<string>uniFEFF</string>
-		<string>cent</string>
-		<string>colonmonetary</string>
-		<string>currency</string>
-		<string>dollar</string>
-		<string>dong</string>
-		<string>Euro</string>
```
burodepeper: I think it is a smart idea if we also include the Euro sign, or in general, all the currency signs we have available that are still actively in use. $, £, ¢ and ¥ are already covered I believe by Latin-1 and Latin-1 Supplement. Or perhaps even just the currency sign block, 20A0-20CF.

source/Hack-Regular.ufo/lib.plist
```diff
@@ -2,3346 +2,324 @@
...
-		<string>uni10ED</string>
-		<string>uni10EE</string>
-		<string>uni10EF</string>
-		<string>uni10F0</string>
-		<string>uni10F1</string>
-		<string>uni10F2</string>
-		<string>uni10F3</string>
-		<string>uni10F4</string>
-		<string>uni10F5</string>
-
```
burodepeper: 315 glyphs counted.

burodepeper: I expected:
- Basic Latin: 94 glyphs
- Latin-1 Supplement: 94 glyphs
- Latin Extended-A: 128 glyphs

For a total of 316 glyphs. With `017F` missing, that seems about right. I haven't checked for typos, but I feel safe to assume that you didn't manually type this list.

burodepeper: If we consider the missing Euro sign, what other obvious things might we have missed?

burodepeper: I think we should also consider adding: General punctuation (2000-206F) for fancy quotes often found in sources that have been created in non text-editor word processors, and also for the en- and em-dash. In my opinion, those characters should be encoded, but you obviously can't rely on that.

chrissimpkins:
> but I feel safe to assume that you didn't manually type this list.

and you would be incorrect in that assumption ;)  I did type the individual name strings, not the XML elements.  XCode is a very good property list editor.  I don't have any approach that labels 1:1 glyph names in our sets with the Unicode values.

chrissimpkins:
Added the currency symbol / general punctuation suggestions to the TODO list.  It is a balance between what will be used in a monospaced font and the size of the final files.  Users will always have the full sets if they need extended symbols.  Out of interest, do web fonts replace individual glyphs with a fallback font in the CSS if any glyph in the entire formatted body of text is missing in the set, or is it only the individual missing glyph(s) that is replaced?

source/Hack-Regular.ufo/lib.plist
chrissimpkins: > 315 glyphs counted.

ty!

chrissimpkins:
Shall we build some sets and compare final file sizes for woff and woff2 sets?  No idea what kind of overhead there is in the metadata, but with starting sizes for whole sets at ~150kb (woff) and ~100kb (woff2), we should be down in the vicinity of 30kb and 20kb, respectively.  Data that I have is "average" web page size at present is 2MB.  Web fonts take advantage of caching in the browsers so only first page view impact.  Certainly not my area, but to me seems like we have room to add more glyphs if felt to be important.

chrissimpkins:
> and you would be incorrect in that assumption

see how much work your suggestions lay on me :)

burodepeper:
> It is a balance between what will be used in a monospaced font and the size of the final files.

I think the € is a must-have, as are the fancy quotes. I come across both of those myself on a regular basis. Extending that to their (relatively small) groups seems to cover similar use-cases.

> Out of interest, do web fonts replace individual glyphs with a fallback font in the CSS if any glyph in the entire formatted body of text is missing in the set, or is it only the individual missing glyph(s) that is replaced?

As far as I know, web fonts are treated like any other font, they're just loaded off of the web. That would mean that individual missing glyphs will be replaced by whatever the software or OS uses as fallback, and at that, it will _not_ use a monospaced fallback, if monospace isn't specified in the CSS as fallback in the font-family. We need to check that too.

> Shall we build some sets and compare final file sizes for woff and woff2 sets? No idea what kind of overhead there is in the metadata, but with starting sizes for whole sets at ~150kb (woff) and ~100kb (woff2), we should be down in the vicinity of 30kb and 20kb, respectively.

Sure. At those sizes, the biggest factor is the latency setting up a connection to the server/CDN. And that is a good thing, because then it is no longer 'our problem' ;)

burodepeper:
> see how much work your suggestions lay on me :)

Perhaps it is revenge ;)

chrissimpkins:
> Perhaps it is revenge ;)

I have no doubt...

chrissimpkins:
Will try to build them tonight and will update here when we have more information.  

I'm good with all suggestions about expansion of sets that you had above.

chrissimpkins:
@anthrotype Cosimo, I modified our lib.plist files in order to add the Glyphs subset keys (for our web font builds) and during this lib.plist cleanup, I removed the public.glyphOrder key/values.  This caused our fontmake build to fail.

Start line for the fontmake exception:
https://travis-ci.org/source-foundry/Hack/jobs/269423508#L4217

My understanding is that lib.plist is supposed to be an optional file in the UFO spec.  I wanted to see if

1) this lib.plist data is considered mandatory for fontmake builds
2) you might consider converting to contents.plist (mandatory file in UFO spec) for the glyph lists necessary for builds from UFO source

We are struggling with how to manage the editor dependent lib.plist data that is unnecessary in the remote git repository in a collaborative typeface development project.  Ideally, everyone keeps their editor dependent data locally where they need it and builds occur from data in mandatory files that do not rely on the editor in use by the developer who commits.

The changes that occur in these files are not terribly important, but they do require review prior to merge and with thousands of lines that can change it makes for unnecessary work for the source review team.

anthrotype:
Right, it should be optional, although the values prefixed with `public.*` would be good to keep.
Do you mind opening an issue on fontmake's repo?

chrissimpkins:
Here is where we are with full set vs. current ASCII / Latin-1 / Latin Extended A web fonts

#### Full set

full set

#### Subset

subsets

chrissimpkins:
For some reason, the use of the lib.plist files with the subset data fields in them broke the ttf builds *without the use of the `--subset` flag* in fontmake.  I hacked together a temporary source directory approach for subsetting that involves a copy of source directories to temp dir, replacement of the lib.plist without subsetting data with the lib.plist that contains subsetting data, and then subset builds.  

Not pretty but it works...  The temporary source directories are cleaned up at the end of the process and recreated with every new subset build.

As of https://github.com/source-foundry/Hack/pull/291/commits/8895e7e9497df92e34e3465e50f409bb2b80644f, you can build each file type individually:

```
$ make ttf
$ make woff
$ make woff2
$ make subsets
```
or build all of them in one fell swoop:

```
$ make
```

All above require previous dependency installs.  If they are not installed yet, you can install all depdencies and build all release files with:

```
$ make build-with-dependencies
```

chrissimpkins:
The subsets seem to pull in additional glyphs including the likes of superscripts/subscripts/all fractions, though they are not all specified in the lib.plist subset data.  Appear to be one short in bold and bold italic sets c/w regular and italic sets.  Will need to go through them glyph by glyph to confirm that we are capturing what we intend to in the subsets.

Let me know based upon current file sizes (https://github.com/source-foundry/Hack/pull/291#issuecomment-325848444) whether you still feel that we should add the additional sets that you recommended above.

burodepeper:
> Let me know based upon current file sizes (#291 (comment)) whether you still feel that we should add the additional sets that you recommended above.

Sure, those numbers are good!

chrissimpkins:
Currency glyphs added to web font subsets https://github.com/source-foundry/Hack/pull/291/commits/869c0d1d8476b78003fb9db708487ccb34617f51

General punctuation glyphs added to web font subsets https://github.com/source-foundry/Hack/pull/291/commits/c750713fe0e5e8d18c95e8802daff05bd776dd79

chrissimpkins:
Minimal impact on file sizes with addition of new glyphs to subsets

web 2017-08-30 21-10-19

chrissimpkins:
Once the glyph subset review is complete, this branch is ready to merge with `dev-build-scripts` from my standpoint.  Let me know if there is anything further that I can do here.

chrissimpkins:
@burodepeper any issues with merge of this to `dev-build-scripts`? I added the woff/woff2 subset glyph reviews to the TODO list in #286

burodepeper:
@chrissimpkins I haven't had time to proof yet. Have you ever had issues when building the webfonts with missing or visually distorted glyphs? If you haven't, I suggest you merge. If you have, I'll need to find some time to have a look at it.

chrissimpkins:
This will be the first time that we have used this subsetting approach to create the web fonts so we have no prior experience with it.  sfnt2woff-zopfli and woff2 are two established tools for these conversions so I am less concerned with the final renders as with the contents of the subsets themselves.  I just haven't had the time to go through them to confirm that we are putting together the glyph sets that we think that we are.  If I get to it before you do I will let you know.

OK with a merge to `dev-build-scripts` for now and close of this PR so that I can continue to base the rest of the build scripting work on the progress here without a new branch?  We have an IR open for the glyph set contents that is part of the v3.0 milestone.  It won't be released until all tasks are complete.  If you happen to decide to tackle this issue, let me know when you are ready and I will confirm that the web font subsets are being built as they were at this stage.  I don't intend to change the subsetting approach in the scripts *unless we find problems with the current approach*.

burodepeper:
Of course, merge the hell away ;)

chrissimpkins:
:+1:

chrissimpkins:
Also, if you haven't been able to set up the development / build tools yet, you can create the web fonts from any local ttf builds with this OS X tool:

https://github.com/source-foundry/Woffle

It uses the same approach that we are using here.

chrissimpkins:
merged

PR #312, opened 2017-09-16
Declare success_indicator as a local variable, not a GLOBAL CONSTANT

cclauss:
Declare success_indicator as a local variable, not a GLOBAL CONSTANT.
* By convention, values that may change at runtime should not be in UPPERCASE.
* A global variable can be _read from_ either with or without a `global` statement
* A global variable can be _written to_ only when preceded by a `global` statement

```
>>> SUCCESS_INDICATOR = 0
>>> def read_from_global():
...     print(SUCCESS_INDICATOR)
...
>>> def write_to_declared__global():
...     global SUCCESS_INDICATOR
...     SUCCESS_INDICATOR += 1
...
>>> def write_to_undeclared__global():
...     SUCCESS_INDICATOR += 1
...
>>> read_from_global()
0
>>> write_to_declared__global()
>>> read_from_global()
1
>>> write_to_undeclared__global()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in write_to_undeclared__global
UnboundLocalError: local variable 'SUCCESS_INDICATOR' referenced before assignment
>>>
```

### DIFF
--- a/postbuild_processing/posthinted_builds/dev-versioner.py
+++ b/postbuild_processing/posthinted_builds/dev-versioner.py
@@ -11,9 +11,9 @@ import sys
 from fontTools import ttLib
 
 VERSION_STRING="Version 2.020;DEV-03192016;"
-SUCCESS_INDICATOR = 0
 
 def main(argv):
+    success_indicator = 0
     for font_variant_path in argv:
         tt = ttLib.TTFont(font_variant_path)
         namerecord_list = tt['name'].__dict__['names']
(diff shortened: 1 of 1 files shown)


chrissimpkins:
Thank you very much for the corrections!  This script is likely going to go away as of the next release but I think that it is worth keeping this commit in case it does happen to stay for any reason.  Can I ask you to rebase this on the dev-build-scripts branch instead of on master?  My attempt to merge this into our development branches eliminated our CONTRIBUTING.md document because you are based on master here.

Also, if you need to modify version strings, we have a new tool for this:

https://github.com/source-foundry/font-v

We will be using font-v for our versioning as of the next release.

Thanks again!

cclauss:
Close in favor of #312.

PR #314, opened 2017-09-16
Declare success_indicator as a local variable, not a GLOBAL CONSTANT

chrisclauss:
Redo #312 on the correct branch.
Declare success_indicator as a local variable, not a GLOBAL CONSTANT.

* By convention, values that may change at runtime should not be in UPPERCASE.
* A global variable can be read from either with or without a global statement
* A global variable can be written to only when preceded by a global statement
```python 
>>> SUCCESS_INDICATOR = 0
>>> def read_from_global():
...     print(SUCCESS_INDICATOR)
...
>>> def write_to_declared__global():
...     global SUCCESS_INDICATOR
...     SUCCESS_INDICATOR += 1
...
>>> def write_to_undeclared__global():
...     SUCCESS_INDICATOR += 1
...
>>> read_from_global()
0
>>> write_to_declared__global()
>>> read_from_global()
1
>>> write_to_undeclared__global()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in write_to_undeclared__global
UnboundLocalError: local variable 'SUCCESS_INDICATOR' referenced before assignment
```

### DIFF
--- a/postbuild_processing/posthinted_builds/dev-versioner.py
+++ b/postbuild_processing/posthinted_builds/dev-versioner.py
@@ -11,9 +11,9 @@ import sys
 from fontTools import ttLib
 
 VERSION_STRING="Version 2.020;DEV-03192016;"
-SUCCESS_INDICATOR = 0
 
 def main(argv):
+    success_indicator = 0
     for font_variant_path in argv:
         tt = ttLib.TTFont(font_variant_path)
         namerecord_list = tt['name'].__dict__['names']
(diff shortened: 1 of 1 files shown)


chrissimpkins:
:+1: we got it!  This branch includes changes that will be coming as of our upcoming v3.0 release.  Thanks much for the contribution!  Added you to the contributors list as well.

cc: @cclauss

PR #320, opened 2017-09-19
Middle dot (U+00B7) fixes for bold and bold italic sets

chrissimpkins:
Based on `dev-build-scripts`

- adds middle dot (U+00B7) to the bold italic set
- modifies the middle dot (U+00B7) in the bold set (reduced size of the dot)

This completes the Latin-1 set for all Hack variants based upon analysis by @jublo in https://github.com/source-foundry/Hack/issues/286

New web font builds are located in the `build/web` directory of the `dev-middledot-patch` branch.  Build v3.000 5c587ef-dev


chrissimpkins:
@burodepeper @jublo merging this as I verified that this glyph was actually missing.  This PR adds it to the bold italic set and reduced the size (to that of all other glyph variants since this is most commonly used to indicate spacing in source code) so that it is the same across all variants.

Issue #43, opened 2015-08-31
Number 1 looks too much like an l letter


LeoNeeson:
The "l" letter is fine, but the "1" number may still lead to confusion, since is similar to the lowercase letter "L" in many fonts. To me, the "1" number must be like this: http://i.imgur.com/pHeeYIJ.png
![improved](https://cloud.githubusercontent.com/assets/13534864/9573729/130c6b3a-4f96-11e5-9f57-8a1443e111ea.png)

chrissimpkins:
Can you let me know what pixel size you are working at and the platform you are using this on?  Might be a hinting/instruction issue if the glyphs begin to look similar.

LeoNeeson:
I'm using the normal display size at 72 PPI "DPI", using a Mac OS system. But this it's not exactly about resolution/system configuration, it's more about design. What I'm trying to explain, is that any newcomer to your font, "may" confuse the number "1" (of your font), with the lowercase letter "L" (in many other fonts). I'm not saying that the "l" and "1" are identical in your font.

But that it's up to you (and the community), since there are a lot of fonts to choose. If you are happy with the current "1" number, great. I've reported this, because IMHO this is a problem to me, but I don't mean it's a problem to everyone. It would be great to hear the opinion of other people to see if they agree on this or not. It's a design choice, not a bug report.

willfroning:
I agree. When you see the "l" and "1" next to each other it's obvious which is which, but having a downward slope to the "nose" would more clearly define it's identity.

chrissimpkins:
Noted.  Will take a look at this.  Is there a size threshold where you notice that the problem begins?

chrissimpkins:
@LeoNeeson Point well taken.  I ask because I am working on the optimization of the TT instructions and wanted to see if there might be a way to optimize it as is.  To my eye in that image, the stroke height on the 'l' vs. '1' (taller 'l' glyph) and the curved tail vs. flat bottom serif look distinct.  The problem is that these nuances tend to disappear as we challenge the glyphs to fit the grid at smaller pixel sizes. Height differences generally disappear and curves may be rendered as flat in which case, it would look exactly like the 1.  I will take a look into this in more detail.

Let's leave this open for more discussion.

chrissimpkins:
Queued up to get started on this.  Will post some design examples in this thread if you are willing to participate as we improve on the shape.

LeoNeeson:
Okay, I'll be here to give my opinion.

[...]

PR #347, opened 2017-10-29
Numeral one (U+0031) re-design to address legibility vs. lowercase l (U+006c)

chrissimpkins:
Modifications to the numeral 1 design to address legibility issues with the lowercase l (issue #43)

Includes the following changes to the glyph design in the `dev-one-design` branch:

- increased angle of top stroke relative to horizontal
- increased length of top stroke
- decreased width of the top stroke (bold and bold italic sets) - uses same dimensions as the left angled stroke in the numeral 4 in these sets

See images in #43 thread for the new design.

Please feel free to modify the working branch with design changes to address any concerns that you have.  We will include modifications in this PR for further discussion.

Build files for local testing are available in https://github.com/source-foundry/Hack/tree/dev-one-design/build/ttf and carry the version string `
Version 3.001; 379b3ad-dev`.

Propose merge to `dev` branch as new default U+0031 glyphs in the sets for release as of v3.001. 

When the design is acceptable, I will need to evaluate hinting at our design target range of 8 - 14ppem before merge.

TODO:

- [x] examine hints at 8 - 14 ppem regular set
- [x] examine hints at 8 - 14 ppem italic set
- [x] examine hints at 8 - 14 ppem bold set
- [x] examine hints at 8 - 14 ppem bold italic set


chrissimpkins:
@burodepeper if you can point me to your visual diff repo (with a brief set of instructions) I would be happy to link the design diff for these changes as well.

burodepeper:
It looks awfully similar to U+2960...

chrissimpkins:
= Upwards Harpoon with Barb Left from Bar

Assuming our users are not routinely placing harpoons in their code do you have other thoughts about the legibility issue against the glyph of concern.

chrissimpkins:
hints look good across 8-14ppem x all four sets

PR #358, opened 2017-12-06
Add glyph U+2027

jublo:
`U+2027` `hyphenationpoint` is an optical clone of `U+00B7` `periodcentered` and used to separate syllables.

In addition to adding the clone, this PR also removes a duplicate related glyph from the BoldItalic set. Please ensure to remove it also from the corresponding metadata files.

TODO:

- [x] remove glyph121.glif from contents.plist in bold italic set


chrissimpkins:
Let's replace periodcentered.  Syntax highlighters often use this to display spaces in source code.  We had early issues with this in some editors.  Can merge when this is replaced in the sets.

chrissimpkins:
Sorry I may have misunderstood here.  It looks like the periodcentered glyph was removed in this PR in only one variant set (bold italic).  It should be in each of the four variants.  Can you give me a bit more detail on the change here?

jublo:
For BoldItalic, the set contained `uni00B7.glif` _and_ `periodcentered.glif` – so I removed the former. The other sets do not include this glyph as a duplicate.

chrissimpkins:
Ah!  Fantastic. tyvm @jublo!

chrissimpkins:
Added the removal of that glyph from the contents.plist file in the bold italic set as a TODO here.  I will take care of this as part of the merge.

chrissimpkins:
Likely won't get to these before the weekend.  We will plan to include all of these PR in v3.001.

jublo:
grafik grafik

Italics look identical to Regular and Bold respectively.

chrissimpkins:
@jublo merged!

PR #359, opened 2017-12-06
Add glyphs U+2032..U+2033, rename U+2034

jublo:
TODO:

- [x] confirm that "millisecond" glyph is rendered with new name change

### DIFF
--- a/source/Hack-Bold.ufo/glyphs/millisecond.glif
+++ b/source/Hack-Bold.ufo/glyphs/millisecond.glif
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<glyph name="uni2034" format="1">
+<glyph name="millisecond" format="1">
   <advance width="1233"/>
   <unicode hex="2034"/>
   <outline>
@@ -25,7 +25,7 @@
   <lib>
     <dict>
       <key>com.schriftgestaltung.Glyphs.lastChange</key>
-      <string>2017-07-11 17:15:08 +0000</string>
+      <string>2017-12-06 01:36:01 +0000</string>
     </dict>
   </lib>
 </glyph>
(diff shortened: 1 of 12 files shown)


chrissimpkins:
Confirming that we are breaking build because you renamed a *.glif file?

jublo:
@chrissimpkins Can you update the metadata accordingly?

chrissimpkins:
Yes no problem.  ufolint seems to indicate that it is unhappy because a glif file that is named in contents.plist is missing.  It appears that this is the problem.  Is "millisecond" the way that Glyphs names this file by default?

source/Hack-Bold.ufo/glyphs/millisecond.glif
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<glyph name="uni2034" format="1">
+<glyph name="millisecond" format="1">
   <advance width="1233"/>
   <unicode hex="2034"/>
   <outline>
```
chrissimpkins: It looks like each of the "millisecond" glyphs are missing outline definitions.  Are they rendering the glyph for you?

chrissimpkins:
Updated your OP with a TODO for this once we get to a build that works.  Mind checking this locally by replacing the uni2034 in your `contents.plist` files with "millisecond", building, and checking that you can display the glyph with the fonts?

chrissimpkins:
No need to push the contents.plist.  I can make the same change so that we don't reorder it but it would be helpful to know that your changes render.  If they do not we should change the name back to the unicode version.

jublo:
After replacing `uni2034` with `millisecond` in all 4 `contents.plist` files and then compiling, Font Book says this:

grafik

chrissimpkins:
The source you committed here does not have a shape defined for those glyphs.  Are they unstaged changes in your local repo?

jublo:
grafik

grafik

grafik

grafik

chrissimpkins:
Apologies.  For some reason, Github was folding the source files in the diff.  I didn't see the source lines.  The outlines are there.  Should be gtg

chrissimpkins:
Thanks for working on all of these @jublo

chrissimpkins:
Merged!  Thanks @jublo !

PR #403, opened 2018-02-22
Fallback mechanism for the woff/woff2 dependencies

### LINKED ISSUE #395: [build system] look for system-wide installed binaries
A minor, wishlist issue relevent for downstream packaging.

It would be nicer if the build scripts did look if the `sfnt2woff-zopfli-build` and `woff2_compress` binaries are installed system-wide (i.e., in the `$PATH`), before looking in the user's home directory.

As an alternative, it would be nice to be able to override the `SFNTWOFF_BIN` and `WOFF2_BIN` variables with environment variables.

paride:
If sfnt2woff-zopfli and woff2_compress binaries are not found in the
installation directory targeted when the scripts are called with
--install-dependencies, then look if the binaries are installed
system-wide (i.e., in the $PATH).

IR #395.

### DIFF
--- a/build-subsets.sh
+++ b/build-subsets.sh
@@ -135,19 +135,31 @@ fi
 # sfntwoff-zopfli installed
 if ! [ -f "$SFNTWOFF_BIN" ]
 	then
-	    echo "sfnt2woff-zopfli was not found on the path $SFNTWOFF_BIN.  Please install all build dependencies with 'make build-with-dependencies', then attempt your build again." 1>&2
-	    INSTALLFLAG=1
-else
-	echo "sfnt2woff-zopfli executable identified"
+	    if ! which sfnt2woff-zopfli
+	    	then
+	            echo "sfnt2woff-zopfli was not found on the path $SFNTWOFF_BIN.  Please install all build dependencies with 'make build-with-dependencies', then attempt your build again." 1>&2
+	            INSTALLFLAG=1
+	    else
+	    	SFNTWOFF_BIN="sfnt2woff-zopfli"
+	    	echo "sfnt2woff-zopfli executable identified"
+	    fi
+	else
+		echo "sfnt2woff-zopfli executable identified"
 fi
 
-# woff2 installed
+# woff2_compress installed
 if ! [ -f "$WOFF2_BIN" ]
 	then
-	    echo "woff2_compress was not found on the path $WOFF2_BIN.  Please install all build dependencies with 'make build-with-dependencies', then attempt your build again." 1>&2
-	    INSTALLFLAG=1
-else
-	echo "woff2_compress executable identified"
+	    if ! which woff2_compress
+	    	then
+	            echo "woff2_compress was not found on the path $SFNTWOFF_BIN.  Please install all build dependencies with 'make build-with-dependencies', then attempt your build again." 1>&2
+	            INSTALLFLAG=1
+	    else
+	    	WOFF2_BIN="woff2_compress"
+	    	echo "woff2_compress executable identified"
+	    fi
+	else
+		echo "woff2_compress executable identified"
 fi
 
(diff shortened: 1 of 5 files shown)


build-subsets.sh
```diff
@@ -135,19 +135,31 @@ fi
...
-	    INSTALLFLAG=1
-else
-	echo "sfnt2woff-zopfli executable identified"
+	    if ! which sfnt2woff-zopfli
+	    	then
+	            echo "sfnt2woff-zopfli was not found on the path $SFNTWOFF_BIN.  Please install all build dependencies with 'make build-with-dependencies', then attempt your build again." 1>&2
+	            INSTALLFLAG=1
+	    else
+	    	SFNTWOFF_BIN="sfnt2woff-zopfli"
+	    	echo "sfnt2woff-zopfli executable identified"
```
chrissimpkins: Possible to modify the stdout for users to explicitly indicate that they are building with the system installed version of sfnt2woff-zopfli?

build-subsets.sh
```diff
@@ -135,19 +135,31 @@ fi
...
-	    INSTALLFLAG=1
-else
-	echo "woff2_compress executable identified"
+	    if ! which woff2_compress
+	    	then
+	            echo "woff2_compress was not found on the path $SFNTWOFF_BIN.  Please install all build dependencies with 'make build-with-dependencies', then attempt your build again." 1>&2
+	            INSTALLFLAG=1
+	    else
+	    	WOFF2_BIN="woff2_compress"
+	    	echo "woff2_compress executable identified"
```
chrissimpkins: and here for woff2_compress

build-woff.sh
```diff
@@ -93,6 +93,13 @@ if [ "$1" = "--install-dependencies" ]
 		cd "$CUR_DIR" || exit 1
 fi
 
+# If the binary is not found, look if it is installed system-wide
```
chrissimpkins: or perhaps better to introduce it in these blocks.  Somewhere so that it is clear in the standard output trail that a system installed version is being applied rather than the approach used with the build-with-dependencies target.  I will do the same for the Python installs depending upon how we approach them with version pinning.

chrissimpkins:
In light of @anthrotype's excellent suggestion to establish reproducible builds from a Python build dependency standpoint and my desire to make this the case across all build dependencies, let's move this to a separate script with a new make target rather than creating a new build path that is determined by the presence of system wide installs of these web font build dependencies.  I am happy to support this here but we should not make this the default if reproducible builds are to be a goal going forward.  This PR proposal takes us in the opposite direction and is probably best suited for a new make target that is explicitly requested by the builder and is not the project build make target default.

To support this, I need to change the current approach for both Python dependencies and ttfautohint as currently defined in the dependency install scripting.

I am going to postpone the remaining planned design objectives for the v3.003 release and we will push this transition in our build chain as the last set of changes that we will make before we release the existing `dev` branch changes as v3.003.  I will create a branch off of `dev` where we can work on this and will rebase this PR on that branch.

chrissimpkins:
Rebased this PR on the `build-scripting` branch where the build scripting work for outstanding v3.003 milestone IR work will occur.  Added this PR to the v3.003 milestone.

chrissimpkins:
Modified your OP to prevent automated closure of the associated issue report.  I am going to merge this and use it as a basis for the script updates.  This work will likely wind up on a different script path.  Still trying to decide on the approach.  Let's leave this open to continue the work.  I will take care of the necessary changes to support the new make targets and the needs included in the work here will be part of those new targets.

chrissimpkins:
Current work in https://github.com/source-foundry/Hack/pull/416 invalidates this PR.  If you are interested in performing any of the work to support the remaining open build source refactor items that remain open in #416, please open a new PR.  Thanks Paride.

Issue #379, opened 2018-01-26
[Linux] font height inconsistency with Hack 3.001 release


articuluxe:
![hack-3 1_vs_hack-3 0](https://user-images.githubusercontent.com/3206473/35453402-0242ffd0-02c2-11e8-8473-9394a91ccb95.jpg)

I upgraded today to hack 3.1.0.  I install it by cloning the github repo and copying the ttf fonts from build/ttf/*.ttf.  When using the new fonts, I now notice seeming irregularities in the heights of certain glyphs, using the same application (Emacs 25.1.1) on the same Fedora OS (Linux 2.6.32-431.11.2.el6.x86_64) and same window manager (XFCE).  I use "Hack 10"

I'm including a screen-shot here, on the left is the new 3.1.0, on the right the prior (which I believe was 3.0.0, it would have been whatever was available from the repo as of a week or two ago).  Note how the "r"'s on the left are smaller than surrounding letters; "p" also seems small, while "a" seems big.

chrissimpkins:
Thanks for this report @articuluxe.  This appears to be the same issue that was introduced with v3.001 and reported on Windows systems in #377.  We are hard at work on a fix and I am hoping to have something available over the weekend.  In the meantime, revert back to the v3.000 release and we will notify in both this thread and the Windows thread when it is available.  Sorry for the troubles.  We seem to be having some problems with build dependencies for ttfautohint that have changed between releases.

I will update your report title to indicate that this is a Linux specific issue for users who wander by the repo with the same problem before the fix is released.

chrissimpkins:
I updated our release that includes this bug with information for Linux users to use v3.000 until fix is available.

chrissimpkins:
Will continue the discussion about the fix in #377 unless we identify Linux specific problems that are appropriate for this thread.  Please join us there for the conversation and leave this thread open for Linux user notification.  Thanks

chrissimpkins:
Are you available for testing as we push new builds in an attempt to address this issue? I may have something as early as this evening

articuluxe:
I can try but the Linux box I have access to is at my work, so I won’t have access over the weekend.

chrissimpkins:
No worries.  Will post here when they are available.

chrissimpkins:
If anyone on Linux is available to test this issue, here are a new set of builds that may address the problem:

https://github.com/source-foundry/Hack/commit/e480b1bbc44f5271d40e98deccb35acc55d7bc74

[...]

Issue #377, opened 2018-01-23
[Windows] font height inconsistency with Hack 3.001 release


jlanzarotta:
Hello,

First off, I love the font.  However, after updating my Windows machine to v3.001, the font rendering is somewhat strange for Italics and Bold...  Any ideas on how to fix this?  I reboot, and nothing has changed...

Thanks,

![image](https://user-images.githubusercontent.com/5437686/35288902-f511dccc-0033-11e8-9ef1-7ba71f5c5a83.png)

chrissimpkins:
Can you confirm that you installed this with the WIndows installer that includes the new v3.001 release files?

https://github.com/source-foundry/Hack-windows-installer/releases/tag/v1.5.1

texhex:
Please try the installer if you haven't so far. If you did, please post the file``C:\Program Files\Hack Windows Installer\Log-FontData.txt``. That's a basic log we are generating where we maybe are able to find out what's wrong.

jdw1996:
For what it's worth, I saw a similar problem using my custom build of the font (`make ttf`) on my work machine (Windows 10).

chrissimpkins:
We've had a number of issues related to font caching with residual previous builds that lead to rendering errors on Windows.  Try the Windows installer linked above and let us know if that fixes the problem for you.  We have yet to develop something simple for developers who need to install new builds, test, re-install new builds on the Windows platform.

jlanzarotta:
Yeap, used the latest installer.  Here is the file you requested...  Hope it helps...

[Log-FontData.txt](https://github.com/source-foundry/Hack/files/1661119/Log-FontData.txt)

chrissimpkins:
Thanks @jlanzarotta.  Let's see whether @texhex think there is an install issue here.  If not I will have a look at the tables to confirm that we didn't modify something unintentionally.  

@texhex can you confirm that you are seeing this problem on your Win install?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

texhex:
@jlanzarotta From my point of view, the installation is looking normally, there is nothing that is unexpected.

Idea I have to find out what is going on on your machine: 
* Is the program you are using Java based? Could you maybe try to replicate the problem with Notepad++ and does it show the same errors?

[...]

PR #405, opened 2018-02-25
Checksum script for SHA1 hash testing of OpenType tables

chrissimpkins:
@paride @spstarr @fabiangreffrath

This is an approach that might address the issue with testing upstream/downstream dependency divergence during the build process as your build dependency versions begin to change from those here.  This is a general purpose script and will serve for other font projects as well if you happen to maintain anything other than Hack.  My goal was to produce something that gives you a simple YES/NO answer about whether your build with different tooling versions should render the same as fonts built with upstream tooling versions across the entire build toolchain (including dependencies of dependencies).  

This Python script adds the capacity to perform SHA1 checksum generation and testing across specific OT tables in the font binaries and generate checksums for those relevant tables, excluding tables that do not affect renders and are modified in a time dependent fashion (certain tables are time stamped with times and checksum values dependent upon compile time).  My thought is that we can publish a checksum file with these SHA1 hashes across all desktop (*.ttf) and web (*.woff/*woff2) fonts with each release build.  The script makes valid checksum files that can be tested with `sha1sum` on your distros or this script can be used directly with its own `-c` flag to test the values against the data contained in fonts that you build.  It ignores missing file paths specified in the checksum file and returns a non-zero exit status code if it finds checksum values that do not match expected values for existing file paths.

It will look like something like this on your end for desktop fonts at the moment:

```
# produce your own set of XML dumps of select OT tables after your builds to use for checksum tests
$ cd build/ttf
$ [path to]/checksum.py --exclude head --exclude DSIG --stdout --noclean *.ttf > /dev/null

# pull the upstream checksum file (or perhaps this is the location for our upstream writes of the file?)
$ curl -L -O [path to checksum file in upstream repository]

# perform the checksum checks
$ ./checksum.py -c checksum.txt

OR 

$ sha1sum -c checksum.txt

# cleanup
$ rm *.ttx
```

and for web fonts:

```
# produce your own set of XML dumps of select OT tables after your build to use for checksum tests
$ cd build/web/fonts
$ [path to]/checksum.py --exclude head --exclude DSIG --stdout --noclean *.woff *.woff2 > /dev/null

# pull the upstream checksum file (or perhaps this is the location for our upstream writes of the file?)
$ curl -L -O [path to checksum file in upstream repository]

# perform the checksum checks
$ ./checksum.py -c checksum.txt

OR 

$ sha1sum -c checksum.txt

# cleanup
$ rm *.ttx
```

You can then delete all *.ttx files that you generated for the testing above.  Exit status code of 0 means that you are building what we are building upstream with respect to tables that affect rendering.  Exit status code 1 means that you are building tables that differ from the upstream.  This information will be important for the build tooling authors to know in some cases and we would greatly appreciate being made aware of these issues so that we can take this into account when we consider changes to the upstream build versions as well.  Always happy to help troubleshoot differences that you find in your builds. This script and the script that @anthrotype contributed to the repo will provide a solid approach to examine these changes very easily.  That doesn't mean that there will be a simple interpretation of what they mean but identification of changes in the binary files is no longer as time consuming as it once was and we can offer this to downstream repos.

ttx is part of the fontTools library so there is no new dependency.  fontTools and Python3 interpreter are the only dependencies involved here so there is nothing new to approve.  This script will be merged into the Hack repository and distributed with your source pulls.  

Let me know if this works for you and whether there are any modifications to the script that we should make.  If this works, let us know where it is best to store these checksum files.  It would be easiest and most useful upstream to place them in a repository directory so that we can write them out during testing builds in order to git diff the SHA1 hashes for unexpected changes as we make changes here.  If better as a new file in our releases, we can certainly entertain that idea as well.  My plan is to create a new make target for the tasks on our end to create the checksum files.  We can also add a target to test the checksum files once we identify a location for the checksum files.

TODO:

- [ ] add documentation Markdown page for downstream repos to use as a reference for analysis of builds from source with the `checksum.py` and `ttdiff.sh` tools that will be included in this repo

### DIFF
--- a/tools/scripts/checksum/checksum.py
+++ b/tools/scripts/checksum/checksum.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+#----------------------------------------------------------------
+# checksum.py
+#  A SHA1 hash checksum list generator for fonts and fontTools
+#  XML dumps of font OpenType table data
+#
+# Copyright 2018 Christopher Simpkins
+# MIT License
+#
+# Usage: checksum.py (options) [file path 1]...[file path n]
+#
+# Dependencies: Python fontTools library
+#----------------------------------------------------------------
+
+import argparse
+import hashlib
+import os
+import sys
+
+from os.path import basename
+
+from fontTools.ttLib import TTFont
+
+
+def main(filepaths, stdout_write=False, use_ttx=False, include_tables=None, exclude_tables=None, do_not_cleanup=False):
+    checksum_dict = {}
+    for path in filepaths:
+        if not os.path.exists(path):
+            sys.stderr.write("[checksum.py] ERROR: " + path + " is not a valid file path" + os.linesep)
+            sys.exit(1)
+
+        if use_ttx:
+            # append a .ttx extension to existing extension to maintain data about the binary that
+            # was used to generate the .ttx XML dump.  This creates unique checksum path values for
+            # paths that would otherwise not be unique with a file extension replacement with .ttx
+            # An example is woff and woff2 web font files that share the same base file name:
+            #
+            #  coolfont-regular.woff  ==> coolfont-regular.ttx
+            #  coolfont-regular.woff2 ==> coolfont-regular.ttx  (KAPOW! checksum data lost as this would overwrite dict value)
+            temp_ttx_path = path + ".ttx"
+
+            tt = TTFont(path)
+            tt.saveXML(temp_ttx_path, newlinestr="\n", skipTables=exclude_tables, tables=include_tables)
+            checksum_path = temp_ttx_path
+        else:
+            if include_tables is not None:
+                sys.stderr.write("[checksum.py] -i and --include are not supported for font binary filepaths. \
+                    Use these flags for checksums with the --ttx flag.")
+                sys.exit(1)
+            if exclude_tables is not None:
+                sys.stderr.write("[checksum.py] -e and --exclude are not supported for font binary filepaths. \
+                    Use these flags for checksums with the --ttx flag.")
+                sys.exit(1)
+            checksum_path = path
+
+        file_contents = read_binary(checksum_path)
+
+        # store SHA1 hash data and associated file path basename in the checksum_dict dictionary
+        checksum_dict[basename(checksum_path)] = hashlib.sha1(file_contents).hexdigest()
+
+        # remove temp ttx files when present
+        if use_ttx and do_not_cleanup is False:
+            os.remove(temp_ttx_path)
+
+    # generate the checksum list string for writes
+    checksum_out_data = ""
+    for key in checksum_dict.keys():
+        checksum_out_data += checksum_dict[key] + "  " + key + "\n"
+
+    # write to stdout stream or file based upon user request (default = file write)
+    if stdout_write:
+        sys.stdout.write(checksum_out_data)
+    else:
+        checksum_report_filepath = "checksum.txt"
+        with open(checksum_report_filepath, "w") as file:
+            file.write(checksum_out_data)
+
+
+def read_binary(filepath):
+    with open(filepath, mode='rb') as file:
+        return file.read()
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(prog="checksum.py")
+    parser.add_argument("-t", "--ttx", help="Calculate from ttx file", action="store_true")
+    parser.add_argument("-s", "--stdout", help="Write output to stdout stream", action="store_true")
+    parser.add_argument("-n", "--noclean", help="Do not discard *.ttx files used to calculate SHA1 hashes", action="store_true")
+    parser.add_argument("filepaths", nargs="+", help="One or more file paths to font binary files")
+
+    parser.add_argument("-i", "--include", action="append", help="Included OpenType tables for ttx data dump")
+    parser.add_argument("-e", "--exclude", action="append", help="Excluded OpenType tables for ttx data dump")
+
+    args = parser.parse_args(sys.argv[1:])
+
+    main(args.filepaths, stdout_write=args.stdout, use_ttx=args.ttx, do_not_cleanup=args.noclean, include_tables=args.include, exclude_tables=args.exclude)
--- a/tools/scripts/checksum/checksum.py
+++ b/tools/scripts/checksum/checksum.py
@@ -16,6 +16,9 @@
 #
 # Usage: checksum.py (options) [file path 1]...[file path n]
 #
+#   `file path` should be defined as a path to a font file for all use cases except with use of -c/--check.
+#   With the -c/--check option, use one or more file paths to checksum files
+#
 # Options:
 #   -h, --help          Help
 #   -t, --ttx           Calculate SHA1 hash values from ttx dump of XML (default = font binary)
--- a/tools/scripts/checksum/checksum.py
+++ b/tools/scripts/checksum/checksum.py
@@ -1,18 +1,30 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 
-#----------------------------------------------------------------
+#------------------------------------------------------------------------------------------------------
 # checksum.py
 #  A SHA1 hash checksum list generator for fonts and fontTools
-#  XML dumps of font OpenType table data
+#  XML dumps of font OpenType table data + checksum testing
+#  script
 #
 # Copyright 2018 Christopher Simpkins
 # MIT License
 #
+# Dependencies: Python fontTools library
+#
 # Usage: checksum.py (options) [file path 1]...[file path n]
 #
-# Dependencies: Python fontTools library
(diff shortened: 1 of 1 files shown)


chrissimpkins:
Added `--noclean` flag to above examples.  You will want to keep the .ttx files around to perform your tests against, then delete them. The default is automated removal of the *.ttx files.

fabiangreffrath:
Guys, sorry for chiming in so very late to this discussion, but I think there is some misunderstanding with regard to the meaning of the "reproducible builds" concept.

In the context of Debian packaging, "reproducible builds" means that a Debian source package builds exactly the same binary packages on different systems (i.e. at different dates, with different locales, etc.). It does *not* necessarily mean, however, that the binaries match exactly what is provided by upstream.

The means towards this goal, however, are very similar, i.e. do not include build dates in binaries, do not include GIT commits (because the sources that a package is built from are not necessarily built from a GIT repo), so not rely on a locale's sorting order, do not include user name, build path, or whatever. Anyway, if you as upstream decide to include this information in your release builds, that's perfectly fine.

paride:
@chrissimpkins Aiming at reproducible builds would definitely settle any QC issue, but in general it is not something that is very easy to achieve... For example I'm getting different checksums if I run your script on the reference builds included in the git repository and on the ttf files I build with the Debian packages tools:

```
$ tools/scripts/checksum.py -s build/ttf/*.ttf
1f4faa7ee679ad4fca3cc83fbb482dfffb0f4884  Hack-Bold.ttf
9f552fb0cbd90e85d607b6b5b8c565896c222a96  Hack-BoldItalic.ttf
e8ff72c3ee2407082f8c00c596458a5b79a7fb80  Hack-Italic.ttf
93f5682e020d7ac8ec599692974f90adc02c66d8  Hack-Regular.ttf

$ tools/scripts/checksum.py -s /tmp/hack-debian-build/*.ttf
6ab50020ca611ea1cd7d082b02e7bcef2a96eda7  Hack-Bold.ttf
f7a3985ba931a21746a525933685d192e4080cc1  Hack-BoldItalic.ttf
2e34e083f68bcde055166cb9172299d7e23e9014  Hack-Italic.ttf
af37011c3cb5aa57bd765f5f3ff3de2b01cb673d  Hack-Regular.ttf

$ tools/scripts/checksum.py -ts build/ttf/*.ttf
0ae74eb9cb380840126102351c7f674bd2ea32b4  Hack-Bold.ttf.ttx
a2fe492959e39495cb64203f904e78792748e2bb  Hack-BoldItalic.ttf.ttx
4f624626b60e8a2342e4c60db67f575951ef675f  Hack-Italic.ttf.ttx
e3723e6a69f72d6dd4677c55b4518e52782ee188  Hack-Regular.ttf.ttx

$ tools/scripts/checksum.py -ts /tmp/hack-debian-build/*.ttf
fcec2f4165a542795ac4dd896d8a4b7128b23118  Hack-Bold.ttf.ttx
0d268befa9e988073af34b19b6923b3531b6c56f  Hack-BoldItalic.ttf.ttx
3551bbd5979e113c53b3454a4f9d1d2951def84e  Hack-Italic.ttf.ttx
c5733c801e5ca6fd7b549956e9c52a7dd1586642  Hack-Regular.ttf.ttx
```

chrissimpkins:
> It does not necessarily mean, however, that the binaries match exactly what is provided by upstream.

@fabiangreffrath Absolutely.  And we are not attempting to provide a way for you to create reproducible builds on your distro, nor to perfectly replicate the binaries released upstream in this script.  This script is a resource for QA of your builds if you have the need for it.  It allows you to exclude differences in the way that we define our OpenType tables based upon our project-specific needs from the approach that you use to address your project specific needs, while taking into consideration tables that affect glyph renders for your users, so that you have a way to tell that your forced changes to build dependencies might influence what your users see when they install or upgrade the fonts.  You can certainly create an approach where you perform the tests internally against previous Debian package builds as things change in that package.  That would likely be a more ideal approach because you can use your own internal control of a font that renders well on your distro as the basis to perform testing.  In any case, it's there if you need it.  Ignore it if you don't.  I'll simply reiterate my caution about freely bumping versions of all of these build tools over time without manually reviewing the renders if you choose not to use a more automated method to test against the upstream binaries that are always manually reviewed.

chrissimpkins:
@paride Apologies, I forgot about the version string modifications.  You will need to exclude the name table.  The version string that we use will create a difference there.  Try this command:

```
$ [path to]/checksum.py --ttx --exclude head --exclude DSIG --exclude name --stdout *.ttf
```

on both your builds and the upstream builds.

chrissimpkins:
@paride if that differs, please upload one of your fonts and I will take a look at it to see where the other differences occur.

paride:
@chrissimpkins still different. Here is one font I built:

https://send.firefox.com/download/ed8687f551/#7b6heN2YynCq7f6xyIy0CA

chrissimpkins:
@paride can you point to source state at build time for your fonts?  Will need a way to identify where you are in the source.

chrissimpkins:
@paride actually the better approach is for you to run the diff in your environment.  Mind running this diff script that Cosimo contributed on the same variants (i.e., Regular vs. Regular) for upstream vs. downstream builds?

https://github.com/source-foundry/Hack/blob/dev/tools/scripts/diff/ttdiff.sh

chrissimpkins:
By default the ttdiff.sh script will test all OpenType tables so there will definitely be differences.  We can examine the ones that do not fall in the name, head, and DSIG tables.

chrissimpkins:
Also please confirm that you are taking the build all the way through the completion of our entire default make process.  This script will pick up changes in hints (or absence of them if you are testing against unhinted builds).

paride:
The diff is... big: https://clbin.com/8XD8N. The font I build come from the v3.002 git tag, and they are built using your build script without modifications and without errors.

chrissimpkins:
> The diff is... big

Haha! Now you understand my pain :)  Manual review of these diffs is insane.  The XML dumps are tends of thousands of lines and changes to the glyphs (intentional) lead to hundreds/thousands of lines of changes in the text dump.

anthrotype:
The diff is probably the different ttfautohint version 1.7 vs 1.8.1

chrissimpkins:
And actually this gets back to the retrograde comparison issue that I suggested above.  That clearly won't work because glyphs changes and you will want to diff the tables that define the glyphs.

chrissimpkins:
> The diff is probably the different ttfautohint version 1.7 vs 1.8.1

This may be the issue that we ran into with our attempt to bump ttfautohint in v3.001...  Something changed in that version and when we attempted to bump to that and current versions of FreeType and Harfbuzz it created an odd variable height rendering issue in lowercase glyphs.  Still unclear if this was something on our end or on the other end...

chrissimpkins:
@paride Cosimo is correct.  Those are all changes in the hints with the change to ttfautohint v1.8.1.  We ran into problems with that version when used with what were current at the time FreeType and Harfbuzz build dependencies for ttfautohint.  Have you had a look at the lowercase glyphs with those builds?  

If helpful, you can use this tool to rename the fonts to something other than Hack so that you can review them side by side:

https://github.com/chrissimpkins/fontname.py

chrissimpkins:
#377 and #379 is where they were reported

chrissimpkins:
Great example of why this is useful.  For those who do want to use it to checksum against upstream it will provide reassurance that you are 'diffing' against something that someone has looked at so that you know when it could be important to view the renders before you release your package.  I don't have a Debian test VM but do test on Ubuntu (now and since the v3.001 release issues...), Arch, macOS, and Windows 10.  I try to cover as many bases as possible but there will be things that sneak through no matter what we do.  You can't cover all possible renders out there.

chrissimpkins:
And points out an interesting issue for packagers, because I don't know how you will fix these issues.  We have the option to revert build dependencies which, as I understand it, may not be an option for you?  We will not be able to make source changes upstream to influence these issues.

paride:
What I can say is that I totally do not observe the inconsistencies reported in #377 and #379. The font looks good here...

chrissimpkins:
That is the important thing Paride.  These automated approaches can only tell if something differs, not that the difference is of importance.

chrissimpkins:
It also probably means that Werner is working on Debian/Ubuntu :)

paride:
Hack_Debian (upper) vs Hack_Original (lower)
2018-02-25-233855_1485x136_scrot

chrissimpkins:
@paride 👍 

Cosimo pointed out this tool which I intend to start testing for our visual diffs:

https://github.com/googlei18n/fontdiff

Might be of use to you as well.

chrissimpkins:
fixed error in variable name for a stderr message in https://github.com/source-foundry/Hack/pull/405/commits/35a6b098b33efdde1a3cf8419a0d42023254315c

chrissimpkins:
Merging this script to `dev` with the intent for this to be in the master branch as of our next release.  Will not create a checksum file approach with make targets until I establish tests around this script.  It is here for testing on your end if helpful to you.

Issue #414, opened 2018-03-07
Add Python package PEP508 dependency spec and dependency safety check through pipenv (CI testing)


chrissimpkins:
```
$ pipenv check

Checking PEP 508 requirements…
Passed!
Checking installed package safety…
All good!
```

Returns zero exit status code on passed tests.  Add to CI tests

chrissimpkins:
Added

Issue #406, opened 2018-02-26
Pin version of ttfautohint and its build dependencies FreeType and Harfbuzz


chrissimpkins:
Current scripting supports use of a system installed version of ttfautohint.  Remove this option and support a defined version only in the default make target approach to builds.  

TODO: 

- [x] modify build scripting to eliminate detection of a system installed version of ttfautohint when present.  Only use the local, non-PATH installed versions of ttfautohint and its build dependencies.

chrissimpkins:
@paride in light of the approach that you propose in https://github.com/source-foundry/Hack/pull/403, would it be useful to have make targets that maintain detection of system installed versions of this dependency?

chrissimpkins:
We can consider a complete set of system installed build dep make targets in order to support the approach across the board that you are using in #403 if that is helpful.  We are trying to balance our desire to pin versions with your use of the build tooling here and inability to pin in the same fashion that we will use upstream.

paride:
I would be completely fine with separate build targets that assume that all the needed binaries are available system wide.

chrissimpkins:
I will make that happen.  Let's separate the build process into one that is fully pinned across the board and one that defaults to whatever is installed on a given system.  This achieves flexibility for those who want it and understand the possible consequences of building with whatever they have along with the capacity for us to pin and control this for ourselves upstream.  

Will be working on this over the upcoming week.  Will need to revise your open PR as well.

paride:
Feel free to close it and redo it in the way you prefer.

chrissimpkins:
It should only require removal of the previous local path checks and default to your source with a different script path. I will get started on the work on pinned versions to see what we need there.  If you have time to write the shell script(s) that behave how you need it would be helpful.  If not I will get to it as we make these changes.

chrissimpkins:
This and other build tooling changes that were previously marked for v3.003 will be backwards incompatible changes vs. the current default build process and may break builds for those who are compiling fonts from source.  I am going to push the current `dev` branch changes as v3.003 over the upcoming week.  The work on our build tooling version defaults across the existing shell scripts will be released as new major version v4.000 to appropriately indicate this backwards incompatibility for anyone building from source out there.  I will make this a build tooling modification only release with an anticipated release timeframe of 3 - 4 weeks after v3.003 is available (i.e., about one month from now). 

Design changes that were planned for v3.004 will now be included in a v4.x release.  Design change milestones will be updated after v4.000 is out the door.

Issue #401, opened 2018-02-21
Set SOURCE_DATE_EPOCH relative to git commit time to create reproducible timestamps in the OT tables


chrissimpkins:
This will permit us to create reproducible OT table build structure at a given git commit.  This opens up the possibility to use text diffs for comparison across build tooling versions, Python interpreter versions, different platforms, etc.

Discussion with @anthrotype who suggested this approach begins at this post:

https://github.com/source-foundry/Hack/issues/398#issuecomment-367432102

chrissimpkins:
With the latest release of the fontmake compiler and its ufo2ft dependency, we were able to get to fully reproducible font builds *at every git commit* using the git commit Unix date/time of the checked out commit at build time.

We accomplished this by setting the `SOURCE_DATE_EPOCH` environment variable to the git commit Unix timestamp value with the following in the shell scripts:

```
export SOURCE_DATE_EPOCH=$(git show -s --format=%ct HEAD)
```

and modifying all fontTools based post-compile writes to the binaries to turn off recalculation of timestamps on new file writes.  This was necessary in the DSIG table write and the fstype fix scripts.

The modification was to TTFont object instantiation with the following:

```python
font = ttLib.TTFont(file=path, recalcTimestamp=False)
```

Notably, the TTFA table write by ttfautohint does not appear to update the checksums/timestamps in the binary and no modifications were necessary there.

These changes are now in the `dev` branch and will be included as part of the v4.000 tagged release.  You can verify the changes by checking out a git commit after 752fbdb76dd9148adfe7e00158bb40f5d59e9ab1 and comparing SHA1 checksum values across files that are built at different times.

👍 Big thanks to Cosimo, Khaled, and gang for the changes in the compilers that supported this.  This is extremely helpful.

cc: @anthrotype @khaledhosny

chrissimpkins:
@paride @fabiangreffrath the changes mentioned in https://github.com/source-foundry/Hack/issues/401#issuecomment-397102332 will get you to Debian-specific reproducible/deterministic builds at every git commit.  This was a topic that Fabian raised a few months ago.  I am hoping to push all of the build system changes (in v4.000) that we discussed over the last few months by late June/early July.  Almost there...

Issue #400, opened 2018-02-21
Pin Python build dependencies to defined versions


chrissimpkins:
Establish reproducible builds that permit OT table dumps with ttx to be diff'd by git commit.

Suggestion by @anthrotype to create requirements.txt file:

https://github.com/source-foundry/Hack/issues/398#issuecomment-367435437

anthrotype:
If you don’t mind the noise in your inbox pyup.io runs a bit that will automatically send PRs to update a requirements.txt files in your repo whenever any package mentioned there has a new version published in PyPI.

anthrotype:
(a *bot*, not a bit)

chrissimpkins:
Good service.  It serves the changelogs as part of the PR too when they are available through its automated changelog parsing method.  Good idea.  Thanks Cosimo!

anthrotype:
Yeah, it’s good. You may want to configure it’s frequency to say once a week or so because it gets pretty quickly annoying. Check the configuration file in fontmake repo for example

chrissimpkins:
Convert to the following approach in build scripting to install all Python dependencies:

```
pip install -r /path/to/requirements.txt
```

and eliminate all `pip install [project]` commands in the scripts.  This means that we should be documenting the best approach for builds to be within a virtualenv.

chrissimpkins:
@anthrotype I could use your feedback about how to accomplish this on my development machine.  Do you suggest that I create a virtualenv, activate & install requirements.txt dependencies in the venv, build from this, delete the venv after build, and create new one for every new build?  Homebrew versions of the Python interpreter will be rolling on my development system and this is desired for other reasons on that machine.  Possible to swap out the Python 3 interpreter in an existing venv so that the libraries and executables do not need to be reinstalled each time this changes?  Is the Python interpreter completely isolated in the venv and does it break if the venv is defined with Homebrew installed Python where cleanups remove previous versions of the installed Pythons?

anthrotype:
> Possible to swap out the Python 3 interpreter in an existing venv so that the libraries and executables do not need to be reinstalled each time this changes?

anthrotype:
> Possible to swap out the Python 3 interpreter in an existing venv so that the libraries and executables do not need to be reinstalled each time this changes?

no, each venv is connected to one specific python interpreter, cannot be shared. You need to create two different ones.

[...]

PR #416, opened 2018-03-15
Refactor build scripting to support Python 3 only builds, pinned build dependency versions, optional system installed build dependencies

chrissimpkins:
This PR updates the build shell scripts and Makefile to support:

- Python 3 only builds (pipenv virtualenv based pinned Python version with Pipfile/Pipfile.lock = default & system installed Python 3 = optional)
- pinned build dependency versions (with default `make` command)
   - Python build dependencies - pipenv virtualenv based builds by default
   - compiled dependencies - pinned version local installs off of system PATH by default
- optional rolling system PATH installed Python and compiled build dependencies (with new make targets) - based upon request from @paride for Debian package that is compiled from source

TODO:

- [x] add support for ttf builds in pipenv isolated Python environment with pinned Python version, pinned Python build dependency versions, and pinned compiled build dependencies by default
- [x] add support for ttf builds with system installed Python dependencies and system installed compiled build dependencies as optional approach through different make targets
- [x] modify ttf build test for executable file so that it does not run during system installed dependency builds (https://github.com/source-foundry/Hack/pull/416#issuecomment-373313097)
- [x] add support for woff local install with new shell script and make target
- [x] add support for woff compiles from local pinned version install of sfnt2woff-zopfli executable
- [x] add support for woff compiles from system installed versions of sfnt2woff-zopfli executable
- [x] add support for woff2 local install with new shell script and make target
- [x] add support for woff2 compiles from local pinned version install of woff2_compress executable
- [x] add support for woff2 compiles from system installed versions of woff2_compress executable
- [x] add support for woff + woff2 subsets with local installs of build dependencies with new shell script and make target
- [x] update build CI testing with new default build targets 
- [x] add shellcheck shell script linting for new scripts
- [x] remove `from __future__` import statements in Python scripts
- [x] Python source formatting updates to pass flake8 tests
- [x] convert to upstream woff2 now that macOS build bug is fixed
- [x] test conversion of the scripted pipenv build/installs to `pipenv install --ignore-pipfile` for user development boxes (https://github.com/source-foundry/Hack/pull/416#pullrequestreview-104517621)
- [x] add `pipenv check` Python environment dependency tests to CI testing (https://github.com/source-foundry/Hack/issues/414)
- [x] update scripts to pull new woff2 commit for updated brotli dependency

Addresses the following IR:

- https://github.com/source-foundry/Hack/issues/397
- https://github.com/source-foundry/Hack/issues/395
- https://github.com/source-foundry/Hack/issues/400
- https://github.com/source-foundry/Hack/issues/401
- https://github.com/source-foundry/Hack/issues/406
- https://github.com/source-foundry/Hack/issues/414
- https://github.com/source-foundry/Hack/issues/419

and represents the "backwards incompatible" set of build workflow changes that will be released as v4.000.


chrissimpkins:
@paride I've implemented an approach for compiles of the desktop ttf files using either pinned versions of all build dependencies (in a pipenv venv with pinned Py3, pinned Python build dependencies, and pinned compiled dependencies) or system installed versions of Python 3, Python build packages, and compiled packages in this branch.  I've broken the CI tests until all build artifact scripting is complete so ignore the CI testing results for now.  The default approach with pinned versions is working for desktop font builds without issues for me in local testing.  Can I ask you to confirm that the approach to build with system PATH installed build executables and Python 3 interpreter works for you and appears to be an approach that adequately addresses the following IR that you submitted:

- https://github.com/source-foundry/Hack/issues/397
- https://github.com/source-foundry/Hack/issues/395

Confirm that all necessary build dependencies are installed on PATH.  The make target for the desktop (*.ttf) testing on your end will then be:

```
$ make ttf-system
```

Once all of the builds are refactored to support this new approach, you will be able to use:

```
$ make build-system
```

to install *.ttf, *.woff, *.woff2 (including all subset files) in a single make target with system installed build dependencies.

This will not address https://github.com/source-foundry/Hack/issues/394 and we will not be making the ttfautohint adjustments to address that IR as part of the v4.000 work.  That will require a fair bit of visual examination to push ttfautohint to v1.8.1 since we have had issues with that version in the past.  Will be work for down the road.  The flag is still supported in v1.8.1 of ttfautohint.

cc @anthrotype - this implements a number of your suggestions in various open IR if you are interested.  I ended up going with a pipenv approach and everything is pinned (Python interpreter, Python packages, compiled projects) in the default build.

paride:
https://github.com/source-foundry/Hack/pull/416/files#diff-4dcc5008412e33290f6089c60c2e1579R68 fails, as `$TTFAH` does not contain the path to the file.

After bypassing that test the `ttf-system` target worked fine for me.

chrissimpkins:
Woops.  That shouldn't be tested when you use those make targets.  I changed it in the woff builds.  Will change it in that script too.  Thanks Paride.  Confirming that this is what you were looking to achieve in those IR?

paride:
Yes, this solves both the issues.

As a side note, this part of the script:

```
# create directory if it does not exist
# (occurs with git + empty directories)
if ! [ -d build/ttf ]; then
›       mkdir build/ttf
fi
```
makes me think that you assume that the built fonts are not part of the git repository. But they actually are (you are not .gitignoring them). This is not a problem at all, of course, I just want to be sure this is intended.

chrissimpkins:
That was carried over from the old scripts.  The comment that I left in the code suggests that I once had a problem when the dir was empty.  Don't recall the issue.  Will have a look at it and can eliminate if I can't reproduce the previous problem.  Thanks Paride.

chrissimpkins:
Will add system installed dependency build testing to our CI tests to confirm that future changes to the scripts do not break this support.  Added as a TODO in the OP.

chrissimpkins:
Broken builds of ttf files with system installed dependencies fixed in https://github.com/source-foundry/Hack/pull/416/commits/353455d38eff8e9b72e7cf8056e9bbd86c81cc53

chrissimpkins:
All steps are now implemented.

I can confirm that `make` builds *.ttf, *.woff full + subsets, *.woff2 full + subsets using pipenv venv on my local macOS machine.  Build and script linting tests updated and now passing on Semaphore CI Linux boxes with the same approach.  We seem to be gtg with the new default build approach.

If someone would be willing to give the full system installed build dependency builds a shot with:

```
$ make build-system
```

it would be greatly appreciated!

paride:
It seems that one target is still missing:
```
make: *** No rule to make target 'subsets-system', needed by 'webfonts-system'.  Stop.
```

chrissimpkins:
Sorry! Just pushed it.  I *think* this should be correct now :)

chrissimpkins:
@anthrotype I think that the only missing version pin is now in the brotli git submodule of the woff2 repository.  I don't have experience with pinning submodules to a specific version tag but based upon some reading, it appears that I can use the following to achieve this? 

```sh
git clone --recursive https://github.com/google/woff2.git
cd woff2/brotli
# checkout the desired brotli tag
git checkout v1.0.3
cd ..
# checkout the woff2 tag
git checkout v1.0.2
make clean all
```  

Will this achieve the desired effect of pinning both the woff2 source and its brotli dependency at those version git tags?

Pipfile
```diff
@@ -0,0 +1,20 @@
+[[source]]
+
+url = "https://pypi.python.org/simple"
+verify_ssl = true
+name = "pypi"
+
+
+[packages]
+
+fonttools = "*"
```
anthrotype: you don't need to specify fonttools in addition to fontmake, because the former is a dependency of the latter.

build-pipenv.sh
```diff
@@ -0,0 +1,54 @@
...
+BUILD_PYTHON_VERSION="3.6"
+
+if ! which pipenv
+	then
+		echo "Unable to detect a pipenv install.  Please install with 'pip install pipenv' then repeat your build attempt." 1>&2
+		exit 1
+fi
+
+# install fontTools and fontmake build dependencies with pipenv
+pipenv install --python $BUILD_PYTHON_VERSION --ignore-pipfile fontmake fontTools
```
anthrotype: if you checked in the Pipfile (which already states the top-level dependencies), why do you need to duplicate them here on the command line? Shouldn't `pipenv install` just work?

paride:
@chrissimpkins I did try `build-system` and it works fine now!

anthrotype:
> we simply fix the woff2 version and a fixed brotli version comes with it.

right, that's how git submodules work

Pipfile
chrissimpkins: Safe to depend on the fact that this will always be the case?  We have local Python scripts that depend on fontTools and will not pull it in with a pip install. pipenv simply ignores it since it was pulled in with fontmake.

anthrotype: ok, if you've scripts that directly depend on fonttools then it makes sense to have it there.

build-pipenv.sh
chrissimpkins: Believe that you are correct.  We should be able to simply do this now that I have defined the venv in my development repo with the above command:

```
pipenv install --ignore-pipfile
```

My understanding is that without the `--ignore-pipfile` flag, it defaults to the specification in the Pipfile rather than Pipfile.lock (i.e. `ignore pipfile.lock is default).  Kenneth's docs state that Pipfile should keep versions unpinned.  Unclear why this is the approach.  It seems to me that if a Pipfile.lock is present, the behavior should be to default to pinned versions defined there, else use Pipfile spec which is defined by the versions contained in the current versions of the dependencies at any given time.  

I will test in a cloned repo so that I can replicate what a user's development repo will see when they go through the build.

chrissimpkins:
@paride 

> I did try build-system and it works fine now!

Excellent!  Are we at a place that addresses all of the recommendations in your issue reports and needs in your package?

chrissimpkins:
@anthrotype 

> right, that's how git submodules work

OK, I must have misunderstood the SO article that I read.  It discussed checking out specific version tags in the submodule but perhaps that was to modify the version specified in the top level git repo?

For my own knowledge, is it possible to do what I suggested should you want to modify the git submodule relative to that defined in the top level git repo?  I know, likely not advisable.  Trying to understand how these work.  Is it simply like an semi-isolated (other than top level project defines the commit where it is fixed) embedded git repository that can be manipulated as with a standard git repo when you are located within one of its directories/sub-directories?

chrissimpkins:
@paride 

I also removed the `from __future__ import` statement in https://github.com/source-foundry/Hack/pull/416/commits/e20d78677fe0a17b2fc12bdeebb15caf891c2690.  This was a request in your Python 3 transition IR.  I believe that this is the only location where this is found in local scripts in the project.

chrissimpkins:
And in case you missed it, when you build-system now, you are building with python3.  This is explicit in the script.  Both approaches use python3 only.

anthrotype:
> Is it simply like an semi-isolated (other than top level project defines the commit where it is fixed) embedded git repository that can be manipulated as with a standard git repo when you are located within one of its directories/sub-directories?

pretty much, yes.

> it does not appear that woff2 is very actively developed

send a PR to update the brotli submodule and I can ping the right people ;)

build-pipenv.sh
anthrotype: hm.. what's the point of you checking in the Pipfile.lock if you're going to ignore it?

anthrotype: ah, ok `--ignore-pipfile` option

> Ignore Pipfile when installing, using the Pipfile.lock.

anthrotype: > Kenneth's docs state that Pipfile should keep versions unpinned.

correct. Because the abstract Pipfile is what generates the concrete Pipfile.lock (a bit like setup.py vs requirements.txt)

If you did `pyenv install` without the `--ignore-pipfile` the Pipfile.lock would probably be updated with the current versions? I'm not a pipenv users so I'm not familiar with it.

paride:
@chrissimpkins It seems to me that all my packaging related issues are now solved, and Hack 4 will be packaged without the need to patch anything. Thanks a lot!

chrissimpkins:
@paride 

> It seems to me that all my packaging related issues are now solved, and Hack 4 will be packaged without the need to patch anything. Thanks a lot!

Great to hear that!  Happy to help do our part to make what you do easier.  

@fabiangreffrath you mentioned that you would like to get to reproducible builds in the Debian package during our discussions about these changes.  See the changes that we made here, in part based upon requests from Paride.  There will be new optional make targets available as of v4.0 that allow you to build fonts with system PATH installed versions of all build dependencies in this project.  We also moved to use whatever is defined as `python3` as the python interpreter for all Python packages and local Python scripts.  If you have any feedback on the approach, particularly if you suggest changes to help you support the desire for reproducible builds better, please weigh in.

chrissimpkins:
For all other stakeholders in the build, now is your chance to weigh in on the changes.  Once we push this I would like to avoid further build process breaking changes for some time.  Let's have the conversations now and avoid the need to break the build in the near term.  

My next planned major version release after these changes will include the Python bindings for brotli and zopfli so that we can transition to the use of Python tools to create our web fonts.  This will be most important to me for our webfont subsets because it will simplify the process significantly relative to our current fontmake hack to create subsets.  The introduction of these (at least) two new build dependencies will take a bit of time to propagate through the downstream packaging approval processes on Linux distros and I would like to allow projects some time (6-12 months) to achieve this before we make this move.  I am hoping to avoid other "backwards incompatible" changes to the build process during this period.

Everything is open to feedback and debate.  Make target names, transition to Python 3 default, other build needs that exist out there, etc.  Let us know.

chrissimpkins:
Converted to upstream woff2 in https://github.com/source-foundry/Hack/pull/416/commits/96a35cb528c3784e80326f466a99042be0f46938.  This is pinned at v1.0.2 (woff2) and includes v1.0.1 of brotli.

Submitted PR https://github.com/google/woff2/pull/110 to woff2 project in order to bump brotli to v1.0.3.  Changelog states that compression ratios improved with 1.0.2 --> 1.0.3 changes.  We will see if this leads to smaller woff2 web fonts.

tools/scripts/install/woff2-compress-build.sh
```diff
@@ -14,6 +14,7 @@
 # The woff2 git clone directory.
 BUILD="$HOME"
 INST="$HOME/woff2"
+WOFF2_VERSION_TAG="v1.0.2"
```
chrissimpkins: location of new definition of woff2 project version (by version git tag)

tools/scripts/install/woff2-compress-build.sh
```diff
@@ -29,13 +30,16 @@ echo "#####"
...
-git clone --recursive https://github.com/source-foundry/woff2.git
+# clone the woff2 repository
+git clone --recursive https://github.com/google/woff2.git
 
 cd "$INST" || exit 1
 
+# checkout desired version tag
+echo " "
+echo "Checking out woff2 version tag $WOFF2_VERSION_TAG"
+git checkout $WOFF2_VERSION_TAG
```
chrissimpkins: woff2 is now checked out at the shell script constant defined git version tag instead of pulling from a downstream repository

build-pipenv.sh
chrissimpkins: > If you did pyenv install without the --ignore-pipfile the Pipfile.lock would probably be updated with the current versions? I'm not a pipenv users so I'm not familiar with it.

Yes, believe that this is correct Cosimo.  Docs state that it is a goal of the project to use current versions of packages when possible for safety (security patch releases) reasons.  You can default to Pipfile.lock specified pinned installs with the above flag as I understand it.

Updated the shell script with your suggestion here and it is working as intended and eliminates forced rebuilds of the entire venv with each font compile.  Thanks for catching this!

chrissimpkins: Will be interested to see how this works once system Py3 bumps to Py3.7 on my development machine.  The intended behavior is for `pyenv` to automatically pull the Py3.6 interpreter and install into the venv when it is not available locally.

chrissimpkins:
As of https://github.com/source-foundry/Hack/commit/bc6a683fba3170e16eb690f45bbce5043e19d10a updated build documentation is available that is current with all changes here.

chrissimpkins:
Updated woff2_compress to use brotli v1.0.3.  Renders with updated build tooling 👍  c/w previous builds.  Pre:post waterfall:

k7nka-image

chrissimpkins:
No significant decrease in the woff2 file sizes with the brotli v1.0.3 updates.

chrissimpkins:
Merged into `dev` branch and will be included in v4.000 release of Hack

PR #432, opened 2018-05-28
[U+0028, U+0029] Made parentheses `(` and `)` rounder

jdw1996:
I made the parentheses rounder in all weights and styles to improve legibility and distinguish them from other types of brackets.


jdw1996:
It seems like something went wrong here.  I think I accidentally opened the pull request directly into master.  I will try again.  Sorry about this.

PR #433, opened 2018-05-28
[U+0028, U+0029] Made parentheses `(` and `)` rounder

jdw1996:
I made the parentheses rounder in all weights and styles to improve legibility and distinguish them from other types of brackets.

---
@chrissimpkins edits below:

TODO

- [x] review shapes in multiple source types
- [x] when we have settled on vertical orientation, review brackets/braces and consider alignment of all with same center positions

Issue #417, opened 2018-03-15
Wider `<` [U+003C] and `>` [U+003E]


jdw1996:
This continues the discussion from issue #120.

Tonight I created wider versions of `<` [U+003C] and `>` [U+003E].  Here's a screenshot to give an idea of what the new versions look like.

![screenshot_2018-03-14_21-32-01](https://user-images.githubusercontent.com/17225098/37439800-3117b160-27d0-11e8-83b5-af634689fac0.png)

I'm going to try them out over the next little while and will post more screenshots if I don't see anything that needs more tweaking.  In case anyone is interested in building a version of the font with these glyphs, I'll attach the source for them.

[new-lt-gt.zip](https://github.com/source-foundry/Hack/files/1813701/new-lt-gt.zip)

jdw1996:
I meant to post some rationale behind the change before, but it slipped my mind.

There are lots of ways that different languages use `<` and `>`:
* Comparison operators (`<`, `>`, `<=`, `>=`)
* Bit-shift operators, and stream operators in C++ (`<<`, `>>`)
* Arrows (e.g. `my_object->call_fn();` in C)
* Angle-brackets in XML and HTML (e.g. `<p></p>`)
* Angle-brackets in templates (as in C++) and generics (as in Java) (e.g. `List<String> los;`)
* Various other operators (e.g. `<*>`, `<$>`, `>>=` in Haskell)
Plus, I'm sure, other uses I don't know about or don't use as much.

For some of these cases, the current, narrow `<` and `>` look good, but they seem a bit cramped for others, especially angle-brackets and bit-shift operators.  This is an attempt to find a middle-ground that looks good in all cases.

chrissimpkins:
Looks good Joseph.  Thanks for this work! Can I ask you to post images of the old design vs. new design so that we can compare?  You don't need to do anything fancy.  Side by side or separate images are fine.  This is the easiest way to understand the changes that you made in the design.

We can explore the source idioms in more detail using our source code corpora.  Code blocks that use the glyphs here are searchable across multiple programming languages with the following:

https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/source-foundry/code-corpora%24+%5C%3E+file:%5C.%28c%7Ch%7Ccc%7Cpy%7Cgo%7Cjava%7Cjs%7Cphp%7Crb%29%24

[...]

PR #434, opened 2018-05-28
[U+003C, U+003E] Made less-than `<` and greater-than `>` wider

jdw1996:
I made the less-than and greater-than symbols wider (i.e. taller) in all weights and styles to improve legibility, especially when they are used in XML-style tags.

### DIFF
--- a/source/Hack-Bold.ufo/glyphs/greater.glif
+++ b/source/Hack-Bold.ufo/glyphs/greater.glif
@@ -4,13 +4,13 @@
   <unicode hex="003E"/>
   <outline>
     <contour>
-      <point x="88" y="358" type="line" name="hr00"/>
-      <point x="883" y="641" type="line" name="iv01"/>
-      <point x="88" y="926" type="line"/>
-      <point x="88" y="1176" type="line" name="av01"/>
+      <point x="88" y="268" type="line" name=""/>
+      <point x="873" y="641" type="line" name="iv01"/>
+      <point x="88" y="1016" type="line"/>
+      <point x="88" y="1276" type="line" name="av01"/>
       <point x="1145" y="760" type="line"/>
       <point x="1145" y="524" type="line"/>
-      <point x="88" y="109" type="line" name="av02"/>
+      <point x="88" y="9" type="line" name="av02"/>
     </contour>
   </outline>
   <lib>
--- a/source/Hack-Bold.ufo/glyphs/less.glif
+++ b/source/Hack-Bold.ufo/glyphs/less.glif
@@ -4,13 +4,13 @@
   <unicode hex="003C"/>
   <outline>
     <contour>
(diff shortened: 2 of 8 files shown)


jdw1996:
This PR addresses issue #417.

chrissimpkins:
Windows users can use this development build installer:

https://github.com/source-foundry/Hack-Test-Win-Installer/releases/tag/v1.2.103

chrissimpkins:
Looking great so far Joseph.  The only thing that I have come across is a possible vertical alignment issue with the `->` idiom.  It looks like it may be the dash that may need to be shifted up slightly because we seem to be good with vertical position for idioms like `>=`, `<=`.  I noticed this in C/C++ source while reviewing the changes in #433.

A collection of common source types is available in https://github.com/source-foundry/code-corpora if you'd like to have a look.  Will have a more detailed look through other languages as we work on the chnages in #433.  These design changes are built into the fonts that are linked in that thread.

jdw1996:
You're right, it does appear that the hyphen is a little lower than the centre of the equals sign.  At the font size I use for coding, I hadn't noticed that before.

chrissimpkins:
Will need to look at that glyph in the context of these changes before we merge.  Will probably warrant some changes there

chrissimpkins:
I think that we *should* be set here after the hyphen vertical position change in f34aae6007d317b500acd1e6020c3b5dc3bc49fa.   I centered the glyph to the horizontal stroke on the plus and to the space in the equal symbol.  You can see an example of the orientation vs. your new gt glyph and the equal symbol in line 79 below:

uv3ey-image

Here is a new build if you'd like to take a look.  Let me know if you are happy with this. 

[hack-f34aae600-dev.tar.gz](https://github.com/source-foundry/Hack/files/2082479/hack-f34aae600-dev.tar.gz)

jdw1996:
Looks good to me, I'm happy with it.  Thanks again for all the work you do!

chrissimpkins:
These changes look great.  Thanks for all of the design work here.  They make a big difference.  I am going to take a look in HTML where these will be very apparent and then will inspect both on Windows over the weekend.  To confirm, you are viewing on Linux?

jdw1996:
Yes, I am viewing on Linux, and they look right to me.

chrissimpkins:
OK.  Just confirming instruction sets are ok.  I won't be seeing that on macOS. Thanks!

chrissimpkins:
Lots of cool short idioms in Haskell to explore these changes (as you pointed out in your IR) :)

chrissimpkins:
Completed review of these shapes.  They look great Joseph. Merging to `dev` branch and we will plan to include these changes in the v4.000 release of Hack.  Thanks so much for all of your work on these designs.  It is greatly appreciated!

PR #427, opened 2018-05-02
Hack v4.000

chrissimpkins:
WIP

Coming soon...

Want to test?  Builds at https://github.com/source-foundry/Hack/pull/427/commits/3a85a7679cf302bdd01456070971db6bbbd3cfb0 in the archive below: 

Click to download: [build-3a85a7679.zip](https://github.com/source-foundry/Hack/files/2498307/build-3a85a7679.zip)

PR #447, opened 2018-08-04
Update to Python 3.7 interpreter for build tooling

chrissimpkins:
This PR also updates all Python tooling in the toolchain to current release versions on PyPI.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,4 +10,4 @@ fontmake = "*"
 [dev-packages]
 
 [requires]
-python_version = "3.6"
+python_version = "3.7"
--- a/Pipfile
+++ b/Pipfile
@@ -10,4 +10,4 @@ fontmake = "*"
 [dev-packages]
 
 [requires]
-python_version = "3.7"
+python_version = "3"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -53,47 +53,100 @@
         },
         "cu2qu": {
             "hashes": [
-                "sha256:33459f01d1f7ac35db39c6ad0e894984249c7e298974e1ac797e18717b5f5e39",
-                "sha256:f03a9d0f697bc4334af9cabc4002bfb1fea988bcfe2bffff46530c26b6ba7991"
+                "sha256:0ab43303bc1e010c285bec278f6a6ffcd395a8a0e9c0443718a0e377ce3e704f",
+                "sha256:10a0b22254f9781909f77ab1b471ed0120b9de81a221e97053a11cf4bbf4db6f",
+                "sha256:11a718a57c3e6595eef0fa917ec6c6867e93781cf594e97e753cc37143044755",
+                "sha256:2260c3c638576945985b6a58de24ee0bf33f20cd2e495d529312eaa648066a5a",
+                "sha256:3b9c218986b9b779cdc0768b6c901243c93ae2f337feeaf1c80a79eb8c2df81b",
+                "sha256:4a997dea5a7a07a3649237cefad57aa1134d18687d6268d84190b03d46669218",
+                "sha256:89c46655ea139568ad2d2ae5778244342eb2f12d19b8d2929a538f2082844dea",
+                "sha256:8a30c375d1de626db52c36cbaa1ffcad3dee4e741f41a124b2adf11ba47c82b8",
+                "sha256:8ffda9ab0dec7f6249a93432c274d62a72b9a0ee6787b8027929ced3d6585fb5",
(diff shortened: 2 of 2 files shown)


chrissimpkins:
This proposal merges to dev branch as part of the v4.000 release work in #427

chrissimpkins:
Just received word from the Semaphore CI team that there is now support for Py 3.7 testing.  Will attempt to fix these tests this week.  This PR has been on hold as we waited for this CI testing support.

chrissimpkins:
Initial attempt didn't fix this.  Still working on it with the Semaphore CI team

chrissimpkins:
Closing in favor of a new PR that only includes a Py37 interpreter upgrade. I bumped the Python build tooling dependencies here with pipenv and there is going to be some troubleshooting required to make that move.  With the Python tooling bump, ttfautohint fails because it looks like there has been a conversion away from builds with glyph production names in at least some cases.

chrissimpkins:
@anthrotype Thanks Cosimo.  I think that the only place where this is required in the build process is at the ttfautohint control instructions files style assignments.  We were using the previous default production names from the fontmake 1.5 compiles there.  This will be simple to update.  Probably simpler than figuring out why pipenv will not permit to me to maintain a deterministic build tool set with a Python interpreter version bump.  For some reason it is bumping versions of fontmake deps when I try this despite requests to leave them be.  :)

chrissimpkins:
@anthrotype is there a simple way to get the new glyph names by unicode code point values (from the previous production names) with fonttools?

chrissimpkins:
or do these come from the GlyphData.xml mapping?

anthrotype:
The uniXXXX production names are generated by ufo2ft postProcessor based on the glyph's unicode values, they don't come from GlyphData.xml. The latter is used by glyphsLib to create the public.postcriptNames mapping, which is then applied by ufo2ft postProcessor, when it is present.

Anyway, as I said, if you were happy with the previous glyph production names, just call fontmake with the `--production-names` command line option, or set `com.github.googlei18n.useProductionNames` to true in all the UFO lib.plist.

```xml
<key>com.github.googlei18n.useProductionNames</key>
<true/>
```

anthrotype:
sorry I misstyped the lib key:

https://github.com/googlei18n/ufo2ft/blob/010bc51f25060366f271a57a674931b6f69d6127/Lib/ufo2ft/constants.py#L4

the correct one is

```
com.github.googlei18n.ufo2ft.useProductionNames
```

chrissimpkins:
@anthrotype I would prefer to convert to default behavior with fontmake and this will require modification of the ttfautohint control instruction files.  I need to change the glyph names in those files to match what is in the binary after a build with fontmake.  Is it safe to assume that this is the glyph name defined in *.glif files after this change in the compile tooling?  If not, I need to know where the unicode : name map is or figure out a way to search for the glyph name by unicode value in the binary.

anthrotype:
The default behaviour is no glyph renaming unless a public.postscriptNames mapping is present in lib.plist.

chrissimpkins:
Thanks Cosimo.  Appreciate all of your help!

chrissimpkins:
Worked fine with the glyph name replacements in the control instructions files.  It looks like there was a transition to `ceil` from `floor` somewhere along the way between the versions of the tooling that we are using.  Huge diffs.  :)

anthrotype:
> It looks like there was a transition to ceil from floor somewhere along the way 

Where do you see these diffs?

chrissimpkins:
ttx dump of fonts built with previous Pipfile.lock and built with new Pipfile.lock.

anthrotype:
can you be more specific? e.g. paste an example of such diffs?

chrissimpkins:
and it looks like it is a change to ceil from round, not ceil from floor. X and Y vals in contour definitions are shifted up by 1 unit c/w previous build.  Occurs infrequently relative to all definitions but there are a large number of the changes.  Somewhere between 900-1000 per variant binary.

anthrotype:
i believe it might be a change introduced in fonttools 3.28:
https://github.com/fonttools/fonttools/releases/tag/3.28.0

> [fixedTools] Added otRound to round floats to nearest integer towards positive Infinity. This is now used where we deal with visual data like X/Y coordinates, advance widths/heights, variation deltas, and similar (#1274, #1248).

https://github.com/fonttools/fonttools/issues/1248
https://github.com/fonttools/fonttools/pull/1274

The change is for the good, I recommend you go on with the update.

chrissimpkins:
kldzq-image

PR #457, opened 2018-10-21
Upgrade Python build tooling

chrissimpkins:
PR defines Py3.7 as the Python interpreter version for builds + increases all Python build tooling versions used in pipenv virtual build environment to current PyPI release versions.

These changes required modification of the glyph names used in the ttfautohint control instruction files as the fontmake compiler respects UFO source file spec'd names by default (instead of replacing with production names by default) in the upgraded version of the tooling defined in the Pipfile.lock here.

Build at https://github.com/source-foundry/Hack/pull/457/commits/ac48920c7e766b284ff7d11fff9ef91b1b568dcc: 
Click to download, then unpack and install: [build-ac48920c7.zip](https://github.com/source-foundry/Hack/files/2498744/build-ac48920c7.zip)

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,4 +10,4 @@ fontmake = "*"
 [dev-packages]
 
 [requires]
-python_version = "3.6"
+python_version = "3.7"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,11 +1,11 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "60fe748ce602e0ad6290ca1db986d09913e7fdefee286a24b92af07cf8d9bb77"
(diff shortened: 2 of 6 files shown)


chrissimpkins:
I reviewed diffs of the binaries across these changes and there are numerous (900-1000 per variant) minor single unit contour changes and a handful (7 or fewer glyph positions) of instruction set assembly changes per variant.  All instruction set changes occurred outside of the Basic Latin set.  Should be gtg.

chrissimpkins:
Please note that if you intend to build fonts from source, are using an earlier version of fontmake and its associated dependencies, and are hinting with ttfautohint using the control instructions files in this repository, these changes may break your build.  ttfautohint will not be able to identify the modified glyph names included in this PR.

PR #468, opened 2019-01-26
Fix path to hack-italic-subset.woff

### LINKED ISSUE #467: Incorrect path to italic WOFF in subset stylesheet
The subset stylesheet points to [`fonts/hack-italic-webfont.woff`](https://github.com/source-foundry/Hack/blob/master/build/web/hack-subset.css.in#L23) instead of the correct `fonts/hack-italic-subset.woff` filename.

jbhannah:
Fixes #467.

### DIFF
--- a/build/web/hack-subset.css
+++ b/build/web/hack-subset.css
@@ -20,7 +20,7 @@
 
 @font-face {
   font-family: 'Hack';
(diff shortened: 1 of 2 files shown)


chrissimpkins:
Will get this merged this week and we should push a patch to npm so that this change is reflected in our web font CDN CSS.  Sorry for the delay!

PR #678, opened 2024-09-25
Update Fedora repository in README.md

JosePMorgado:
Change the Fedora hack-fonts repository to the most up-to-date one ([zawertun/hack-fonts](https://copr.fedorainfracloud.org/coprs/zawertun/hack-fonts/)) in Copr, as the ‘heliocastro/hack-fonts’ repository no longer exists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ We are aware of Hack support in the following package managers (with associated 
 - **Arch Linux**: `ttf-hack`
 - **Chocolatey (Windows)**: `hackfont`
 - **Debian**: `fonts-hack-ttf`
-- **Fedora / CentOS**: `dnf-plugins-core :: heliocastro/hack-fonts :: hack-fonts`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,7 +101,7 @@ We are aware of Hack support in the following package managers (with associated
 - **Arch Linux**: `ttf-hack`
 - **Chocolatey (Windows)**: `hackfont`
 - **Debian**: `fonts-hack-ttf`
-- **Fedora / CentOS**: `dnf-plugins-core :: heliocastro/hack-fonts :: hack-fonts`
+- **Fedora / CentOS**: `dnf-plugins-core :: zawertun/hack-fonts :: hack-fonts`
```
mavit, line 104: Hack is packaged as a part of Fedora.  We don't need to use a third-party repository.

```suggestion
- **Fedora**: `source-foundry-hack-fonts`
```

It is not currently included in EPEL for CentOS (but then, neither is it in that Copr project).

